PR #20, opened 2021-07-27
Calibration plugin

camFoltz:
This pull request restructures the calibration code as well as adds the calibration plugin module.

After installing the branch via pip with `pip install -e <path>/recOrder`, you can run the command `recOrder-napari` in your terminal and it should pull up the plugin window.  It will be easiest to test this PR on automaton or hummingbird, but you can take a look through the code.

It is organized such that the main widget lives in `/recOrder/plugin/widget` and the individual workers live in their respective folders.  These workers are moved onto the threads to preform long tasks.  They typically house most of the execution code.

Let me know if you have any questions about the organization / code


recOrder/plugin/plugin_widget.py
```diff
@@ -0,0 +1,535 @@
...
+        self.calib_worker.finished.connect(self.calib_worker.deleteLater)
+        self.calibration_thread.finished.connect(self.calibration_thread.deleteLater)
+        self.calib_worker.progress_update.connect(self.handle_progress_update)
+        self.calib_worker.extinction_update.connect(self.handle_extinction_update)
+        self.calib_worker.intensity_update.connect(self.handle_plot_update)
+        self.calib_worker.calib_assessment.connect(self.handle_calibration_assessment_update)
+        self.calib_worker.calib_assessment_msg.connect(self.handle_calibration_assessment_msg_update)
+        self.calibration_thread.start()
+        self._disable_buttons()
+        self.calibration_thread.finished.connect(self._enable_buttons)
```
bryantChhun, line 419: There's a certain pattern that might be useful here, but probably requires a refactor in a later PR:

Rather than create QThread every time a calibration call is made, you have a separate class (`QThreadWorker` for example) that either creates the thread there or assigns to a thread in an existing thread pool.  The threadworker can also manage connections and update signals.  The code then looks like:

```
def run_calibration(self):
    self.calib_worker = CalibrationWorker(self, self.calib)
    QThreadWorker.start_thread(self.calib_worker)             # here, the worker will move
    QThreadWorker.connect_signals(self, self.calib_worker)      # connects plugin_widget elements and calib_worker elements through a call to calib_worker.
    ...

```

And this allows you to repeat the same structure for all subsequent threads.

Additionally, you often have an `update` and `finished` property for the "worker" classes.  This could become an interface or similar.  Also, some other actions like `worker.deleteLater` and `thread.deleteLater` can all be handled from the QThreadWorker class.

setup.py
```diff
@@ -62,8 +62,8 @@
         packages=PACKAGES,
         include_package_data=True,
         entry_points={
-            'console_scripts': [
-                'recOrder.reconstruct = recOrder.cli_module:main'
-            ]
+            'console_scripts': ['recOrder.reconstruct = recOrder.cli_module:main',
+                                'recOrder-napari = scripts.launch_napari:main'],
+            'napari.plugin': 'recOrder-Calibration = recOrder.plugin.napari_plugin_entry_point'
```
bryantChhun, line 67: will probably need to rename this if the plugin will include more than just calibration

bryantChhun (commented):
Focused on only the `plugin_widget.py` and the `calibration_workers.py`

I left a comment about the way workers are structured.  Overall I think thread creation, movement, and signal connections can be managed in one location.  You can even create a base class/interface with the common signals that "CalibrationWorker" or "BackgroundCaptureWorker" will inherit.  That would all work to compartmentalize functions better and simplify the code for plugin_widget.

It might be easier to diagram on a whiteboard.  We can discuss more later

recOrder/plugin/plugin_widget.py
```diff
@@ -0,0 +1,535 @@
+from recOrder.calib.Calibration import QLIPP_Calibration
+from pycromanager import Bridge
+from PyQt5 import QtCore
+from PyQt5.QtCore import pyqtSlot, pyqtSignal, QThread
+from qtpy.QtWidgets import QWidget, QFileDialog
```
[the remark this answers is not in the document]
camFoltz: Agreed.

recOrder/plugin/plugin_widget.py
camFoltz: Yeah I have to think about this a bit-- I do think that the code to start / manage threads needs to be consolidated

bryantChhun (changes requested):
Some tests don't pass.

It'd be good to have more docs where there are functions with heavy procedures.  I've noted locations where this could be helpful.

Does `CoreFunctions.py` need to be in `calib` module?  I see it as holding general micro-manager abstractions / control functions.  Snap, get/set lc, etc...

After docs adjustments, I think it's good to go.

camFoltz:
Going to close this PR and make a fresh one since many changes have occurred since this was opened

PR #31, opened 2022-01-08
Calibration/Acquisition Plugin

camFoltz:
After many months of development, the Calibration/Acquisition plugin is ready to be merged into the existing codebase.  There are a lot of changes that come with this PR, in all sectors, to support an efficient plugin.  The summary of the main features are below:

the main plugin can be launched by calling `recOrder` in the command line after installation

**Calibration Plugin**
- This plugin allows the user to use either the 4-state or 5-state scheme to calibration the Meadowlark LC's
- Feedback on calibration quality with assessment of the extinction + attempts at diagnosing if there is misalignment in the hardware components
- Load Calibration feature that will allow a user to load a previous calibration from the `calibration_metadata.txt` file.  Useful for micromanager crashes.
- Mimic of OpenCell features like plotting, background capture display, etc.
- Calculate Extinction button that allows the user to move to another FOV and calculate the extinction there
- logging features

**Acquisition Plugin**
- Ability to snap phase/birefringence images or volumes and save them to disk + view them in napari once they have finished reconstructing
- Data Listening: this button can be pressed before an MDA acquisition and it will reconstruct the data as MM writes it to the disk.  CURRENTLY ONLY WORKS WITH MULTIPAGE TIFF FORMAT AND NO FLUORESCENCE CHANNELS.

**Compute**
- streamlines the compute functions and consolidates the `initialize_reconstructor` function

There may be many more changes that I am missing, but I would encourage you all to give the plugin a shot and let me know how it goes.


camFoltz:
I will have to think about the best way for this PR to be reviewed.  Diving into the code may take a while as it adds ~9k lines to the repo.  I know @ieivanov has been using this on mantis and has been very helpful in bringing up design choices + bugs which has worked well, but working through the actual code of this PR could be quite a large task

mattersoflight:
I enjoyed testing the plugin - it is quite functional. Communication with micro-manager was stable. I could not test the accuracy of calibration today. For functional calibration and on-line acquisition, the issues #43, #40, and #37 are a high priority. In the next session, I will continue with testing the accuracy of calibration and orientation.

lihaoyeh:
I am still in the process of testing recOrder. Here are my current findings:

1.	When specifying "QLIPP" with an output channel of only "Retardance", recOrder asks the user to provide parameters like NA, px size, etc, which may not be needed for just the retardance reconstruction.

2.	Data save name is always None when not entered

3.	When specifying "QLIPP" with an output channel of only "Retardance", we do not need the phase transfer functions. The current plugin will initiate a reconstructor with the phase transfer functions no matter whether the output channel contains phase3D or not. Transfer function computation can be conditional when not used to save processing time.

4.	When doing 2D reconstruction, is there a parameter to select which z-slice is the output plane?

5.	When specifying “QLIPP” and Mode “2D”, the reconstruction throws an error if the output channel includes both Phase2D and 3D stacks like Retardance, Orientation, BF. Creating a warning to point out why the reconstruction fails
may be good to avoid the situation.

6.	When specifying “PhaseFromBF”, Mode “3D”, and Output channel “Phase2D”, a Phase2D is reconstructed.

7.	When specifying “PhaseFromBF”, Mode “2D”, and Output channel “Phase3D”, a Phase3D is reconstructed. This suggests that the output channel will overwrite the specification of the dimension of the overall reconstruction.

8.	In the phase regularization section, It seems that the default is to regularize phase3D and there is no option for phase2D regularization. Do we need to have the option for phase2D regularization?

9.	Playing with various regularization values (from 1e3 – 1e-10) for QLIPP Phase3D reconstruction does not result in any difference in the output reconstruction. This may suggest that the plugin does not supply the changed regularization values to the actual reconstruction functions. 

10.	Switching the Pipeline from “PhaseFromBF” to “QLIPP”, the field of brightfield channel index does not clear by itself. You will need to switch from “PhaseFromBF” to “Fluorescence Deconvolution” to make it become Fluor Channel Index. 

11.	When specifying “Fluorescence Deconvolution” and Mode “2D”, it throws an error of focus_zidx must be specified for 2D deconvolution. However, I did not find any place for this parameter in the plugin.

12.	The HSV display panel does not work on the direct output of the QLIPP offline reconstruction. The output channels are aggregated as a whole instance in napari, which the HSV viewer cannot load individual channels from one instance. When I force to use that panel with the incorrect output format, the plugin crashes. Dragging the output zarr file into napari and then using the panel work fine. 

13.	In the standard HSV visualization, we set saturation to be a constant 1 to maximize the color contrast. In the current panel, we must choose the retardance channel as saturation and adjust the upper limit of the channel to make it constant throughout the FOV. Also, adjusting the contrast bars in the Display panel seems to be slow. 

14.	The preprocessing denoising channel should not suggest anything that is from the post-processing, say phase, retardance, ….

15.	When specifying to denoise the retardance channel with threshould [10] and level [1], the reconstruction throws error about string index out-of-range. The config that creates the error currently lives in `CompMicro/sandbox/config_denoise_error.yml`.

I will continue testing the recOrder. Please see if any of these findings are high-priority and need to be fixed at this iteration. Thanks.

ieivanov:
This branch is the de facto "main" branch. I haven't used the actual "main" branch in a very long time. I suggest we merge that at the end of the week, and continue to fix bugs as we find them

ieivanov (dismissed):

mattersoflight:
We may need to write some scripts while we figure out the bugs in the CLI/config parsing modules for specific use cases involving phase and fluorescence deconvolution. I now agree with @ieivanov that it will be easier to do so once the recOrder's main branch has the latest features. Merging now.

Issue #47, opened 2022-01-19
Better tracking of calibration and reconstruction metadata


ieivanov:
In the calibration plugin it would be useful to include a field for user notes. 

I often run multiple rounds of calibration, sometimes under different settings. Right now, I need to save those under different folders or change the filename of `calibration_metadata.txt` to sometime more meaningful. A Notes field would allow me to jot down something quick about the microscope setup that may not be captured in the metadata. In this case, metadata files should not be overwritten, but instead carry a numeric extension, e.g. `calibration_metadata_9.txt`.

Maybe the same Notes field would be useful in the Acquisition widget too

camFoltz:
I can put a notes section under the advanced tab,  I am hesitant to start crowding the main UI with a lot of features that will only be used by a small percentage of users.  That will allow you to write a note directly into the last-calibrated metadata file.

The only thing I am worried about with incrementing the filenames of the metadata file is that if a user is testing the system and performing many calibrations (most of which don't need to be saved) then the folder can quickly become crowded.  My idea here would be to also add a checkbox in the advanced tab that allows the user to choose whether or not to overwrite calibration metadata files (default will be `overwrite=True`).  What do you think of this @ieivanov ?

I am also worried about the numeric extension not being distinguishable enough for a user to remember which calibration it corresponded to.  We may just have to let that be the fault of the user, but my only other idea is to add a timestamp to the metadata file to help with this.

mattersoflight:
I also agree with @ieivanov  that one needs to track metadata for different calibrations (e.g., I am trying swings of 0.03, 0.05, 0.1). @camFoltz I think calibration_<timestamp> is a good format. There is not need to overwrite metadata files since they are tiny. The timestamp gives you a sense of when you ran calibration e.g. at the start of your session, middle or end.

Notes field is also nice to have. It can be a text field next to "Plot" and "Advanced", it can become a "Notes" field, which gets saved in the JSON whenever metadata is written.
![image](https://user-images.githubusercontent.com/2934183/150592456-d622f3e9-ee41-42b3-bd48-5d89e67e456f.png)

Building on this idea, right now we are saving calibrated parameters for pixel-wise reconstruction of retardance. We should document the phase reconstruction parameters in the same way, so that off-line processing can take that as input (#42). How does the current implementation pull the parameters required for phase reconstruction?

camFoltz:
[...]

Issue #39, opened 2022-01-12
check that necessary parameters for acquisition and analysis are supplied before acquiring data


mattersoflight:
I missed to set the parameters needed for phase reconstruction before starting the acquisition. I received following error *after* the acquisition was completed and before the data was analyzed:
<img width="1163" alt="image" src="https://user-images.githubusercontent.com/2934183/149235299-85c7136e-93fe-4120-83aa-644d0b9abcbb.png">

Is there a module that checks whether necessary parameters are provided before the acquisition starts?

camFoltz:
There is not a specific module, but there is similar code elsewhere that's easily adapted for this.  I agree that this will be a good check to implement before moving on with the acquisition.

camFoltz:
implemented in #31

PR #52, opened 2022-01-25
Fluorescence Deconvolution and PhaseFromBF pipelines

camFoltz:
This branch adds two reconstruction methods to the CLI pipeline and also adds their compute API functions.  The two methods implemented are:

- 2D/3D fluorescence deconvolution (with automatic background calculation)
- 2D/3D Phase from BF deconvolution

This will later be called by the offline reconstruction module.  Pytests have been written to ensure reconstruction performs properly but I assume in a later release that the fluorescence deconvolution will need to be optimized a bit further (especially background calculation).


mattersoflight:
@lihaoyeh can you review this PR? I'll continue with testing #31. Please see this discussion on how to launch reconstruction (#39 ) and how to track metadata (#47).

camFoltz:
Thanks @lihaoyeh !

My recommendation would be to test this either by calling the methods here independently: 

https://github.com/mehta-lab/recOrder/blob/fluor_deconv/recOrder/compute/fluorescence_compute.py 

and reference how they're used in 

https://github.com/mehta-lab/recOrder/blob/fluor_deconv/recOrder/pipelines/fluor_deconv.py 

You can also run each pipeline from the command-line interface by defining a config file from an example [here](https://github.com/mehta-lab/recOrder/blob/fluor_deconv/examples/example_configs/config_example_fluor_deconvolution.yml) or [here](https://github.com/mehta-lab/recOrder/blob/fluor_deconv/examples/example_configs/config_example_phase_from_bf.yml)

and using the command:

`recOrder.reconstruct --config <path_to_config>`

recOrder/compute/qlipp_compute.py
```diff
@@ -258,7 +258,7 @@ def reconstruct_phase2D(S0, recon, method='Tikhonov', reg_p=1e-4, rho=1,
 
     """
 
-    _, phase2D = recon.Phase_recon(np.copy(S0), method=method, reg_p=reg_p, rho=rho, lambda_p=lambda_p, itr=itr, verbose=False)
+    _, phase2D = recon.Phase_recon(np.copy(S0).astype('float'), method=method, reg_p=reg_p, rho=rho, lambda_p=lambda_p, itr=itr, verbose=False)
```
lihaoyeh, line 262: When the input dataset has a non-float type (uint), it will throw an error when normalizing the intensity in the `inten_normalization` function. I have added this conversion in the ` reconstruct_phase2D` and ‘reconstruct_phase3D’ functions in ‘qlipp_compute.py’

recOrder/compute/fluorescence_compute.py
```diff
@@ -0,0 +1,148 @@
...
+    return deconvolved_data
+
+
+def deconvolve_fluorescence_3D(data, reconstructor: fluorescence_microscopy, bg_level, reg=1e-4):
+    """
+    Deconvolve 3D fluorescence volume(s).  Will loop through multiple fluorescence channels
+
+    Parameters
+    ----------
+    data:               (nd-array) 2D data of size (N_fluor, Z, Y, X) or (Z, Y, X)
```
lihaoyeh, line 129: Minor: `reconstruct_phase3D` function receives an array with dimension (Y, X, Z) and `deconv_fluorescence_3D` receives an array with dimension (Z, Y, X). We can either make the input of the `reconstruct_phase3D` as (Z, Y, X) or stay what it is and wait for an update in `waveorder`. It is not a big priority, but I note it as an action needed sometime later.

recOrder/compute/fluorescence_compute.py
```diff
@@ -0,0 +1,148 @@
...
+    reconstructor:      (waveorder.fluorescence_microscopy) Initialized reconstructor object
+    bg_level:           (list or int) list/int of background levels of size [N_Fluor] or 1
+    reg:                (list or int) list/int of regularization parameters of size [N_Fluor] or 1
+
```
lihaoyeh, line 133: The `bg_level` and `reg ` for fluorescence deconvolution need to be a list. We could make the documentation consistent with the accepted data type of the functions.

lihaoyeh:
Hi Cam,

Here are other things I noticed when using the command-line interface to reconstruct phase and deconvolve fluorescence channels:

1. Folder specification: When using ome-tiff, `data_dir` needs to be set to the path of the folder that contains all the ome-tiff files. However, when using Zarr, `data_dir` requires a path pointing to the specific Zarr file. We need to be clear about this. 

2. In the `config_example_phase_from_bf.yml`, the `pre_processing` section is removed, which results in errors in reading the config file. If the `pre_processing` section is not needed when reconstructing phase from BF, we need to make sure the config reader would not ask for this.

3. There are two variables, `mode` and `output_channels`, implying the dimension of the reconstruction. When I accidentally specify `mode: 3D` and 'output_channels: ['phase2D']', I get a 2D phase reconstruction.

4. It seems `QLIPP` and `PhaseFromBF` can be used together with fluorescence deconvolution. When documenting the usage, it would be better to note down the input and output of various use cases. 
For example, 
“QLIPP”: <data> Polarization-resolved + fluorescence (optional) intensity. <output> retardance, orientation, phase, original fluorescence or deconvolved (when `deconvolution` is turned on) fluorescence
“PhaseFromBF”: <data> Defocused brightfield + fluorescence (optional) intensity. <output> phase, original fluorescence or deconvolved (when `deconvolution` is turned on) fluorescence
“FluorDeconv”: <data> Fluorescence intensity. <output> Deconvolved fluorescence

5. If the previous points are resolved, we can consider cleaning up the current config to only present relevant parameters for each pipeline. 

Other than that, I could use the off-line phase reconstruction and fluorescence deconvolution smoothly. I have also added an example notebook for the usage of fluorescence deconvolution in the process of testing the code. I am not sure why these additions make the test fail. Let me know if you have more questions regarding these points. If I break anything when pushing two small commits, let me know as well.

recOrder/compute/qlipp_compute.py
camFoltz: Thanks for catching this!

recOrder/compute/fluorescence_compute.py
camFoltz: Ah thanks for catching this -- I think we can leave this for now as it has proper documentation.  It will be nice if `waveorder` can conform to more of the OME standardization of arrays being (T, C, Z, Y, X).

recOrder/compute/fluorescence_compute.py
camFoltz: Done.

camFoltz:
> 1. Folder specification: When using ome-tiff, `data_dir` needs to be set to the path of the folder that contains all the ome-tiff files. However, when using Zarr, `data_dir` requires a path pointing to the specific Zarr file. We need to be clear about this.

I think I will implement this in the waveOrder PR such that if the user specifies a single ome.tif file it'll correct it to the parent directory.  This will be easier in the long run if the reader can handle the subtle differences.

> 2. In the `config_example_phase_from_bf.yml`, the `pre_processing` section is removed, which results in errors in reading the config file. If the `pre_processing` section is not needed when reconstructing phase from BF, we need to make sure the config reader would not ask for this.

Ah yes I have noticed this bug too -- I'll make sure that the config doesn't expect to see categories that aren't going to be used.

> 3. There are two variables, `mode` and `output_channels`, implying the dimension of the reconstruction. When I accidentally specify `mode: 3D` and 'output_channels: ['phase2D']', I get a 2D phase reconstruction.

Interesting, I'll make sure there is a catch for this.

> 4. It seems `QLIPP` and `PhaseFromBF` can be used together with fluorescence deconvolution. When documenting the usage, it would be better to note down the input and output of various use cases.
>    For example,
>    “QLIPP”:  Polarization-resolved + fluorescence (optional) intensity.  retardance, orientation, phase, original fluorescence or deconvolved (when `deconvolution` is turned on) fluorescence
>    “PhaseFromBF”:  Defocused brightfield + fluorescence (optional) intensity.  phase, original fluorescence or deconvolved (when `deconvolution` is turned on) fluorescence
>    “FluorDeconv”:  Fluorescence intensity.  Deconvolved fluorescence
> 5. If the previous points are resolved, we can consider cleaning up the current config to only present relevant parameters for each pipeline.

Yeah I can make sure the example configs are well documented and have only the necessary parameters required for that pipeline.  I will eventually use this information to update the documentation WIKI as well, but I'll plan to do that after the merge and before the first release.

> Other than that, I could use the off-line phase reconstruction and fluorescence deconvolution smoothly. I have also added an example notebook for the usage of fluorescence deconvolution in the process of testing the code. I am not sure why these additions make the test fail. Let me know if you have more questions regarding these points. If I break anything when pushing two small commits, let me know as well.

Yeah I think the tests not passing is due to a google drive data error, I'll look into this.


Thank you so much for the review, I'll make the changes and then we can merge the branch!

PR #87, opened 2022-04-08
Improve installation

talonchandler:
Main question to answer before merge:

- are we using `pip install -e .` for installation? If yes, then I think we can delete the requirements files...they seem to be unused (I tested a fresh install without them).


talonchandler:
Okay I see that the tests failed because of the absence of the requirements files. I'm looking into this to see why the main package seems to depend on the .cfg file while the tests depend on the requirements files. 

Seems to me like we should shoot for a single package dependency file.

talonchandler:
The tests still fail, but they fail in the same way that they did before 🙂,

I've refactored the package dependencies. We can still install with a one-liner `pip install -e .`, and I've removed the requirements files and moved the testing requirements into the github workflow (there were already a few pip install calls there anyway). 

I like going without a requirements.txt file since having both a requirments.txt and a setup.cfg was confusing to me on my first installs: should I do pip install -e . or `pip install -r requirements.txt`. Now it's clearer and we have fewer dependencies files to keep track of. 

Ready to merge from my perspective.

talonchandler:
Following up on a question about how `pip install` works with `.cfg` files, this stack overflow says that `pip install` calls 
`setup.py` which calls `setuptools` which uses the `.cfg` file. More reading required to understand all of this, but at least it makes sense why this still works w/o any requirements.txt files. 

https://stackoverflow.com/questions/46205648/can-pip-install-from-setup-cfg-as-if-installing-from-a-requirements-file

mattersoflight:
From the python packaging user guide, I [read](https://packaging.python.org/en/latest/discussions/install-requires-vs-requirements/#requirements-files) I read that pip installation looks only in the setup.cfg. 

The guide also makes the point that `install_requires` keyword in setup.cfg and the `requirements.txt` serve different purposes. `install_requires` specifies minimal requirements needed to run the code and `requirements.txt` is simply the list of pip commands that a developer can use to setup the development environment. In our case, the pyqtdesigner should be specified in `requirements.txt`, but not in `install_requires`. 

The installation instructions can be:
1. `pip install -e .` to install the package.
2. `pip install -r requirements.txt` if you want to develop the code.

PR #89, opened 2022-04-14
Dist name to recOrder-napari to avoid PyPI conflict

### LINKED ISSUE #384: Package name for PyPi
Unfortunately` recorder is taken on pypi for a python package name.  We will have to consider other deployment names for recOrder.

We could in theory do something like `napari-recOrder` to follow other napari plugins or something similar.  I am not entirely sure what would be best here.  @mattersoflight

talonchandler:
This is a hard-won attempt to change the distribution name to (`recOrder-napari`) while keeping the module name unchanged (`recOrder`) so that `import recOrder` statements still work. 

If we merge this and upload to PyPI I expect the package to appear as `recOrder-napari`. 

I tested this installation with `import`, the CLI, and the GUI as follows:

With this commit in your working directory
```
# Install fresh environement
$ conda create -n recorder python=3.7
$ conda activate recorder
$ pip install -e .

# Python import
python
>> import recOrder
# Imports correctly
>> quit()

# CLI
$ recOrder.help
# Prints help message

$ pip install napari[all]
$ napari -w recOrder-napari
# Opens napari with recOrder-napari plugin

```

Note: this is a branch from `main`, not `npe2_conversion`. I could not get `npe2_conversion` to work with CLI or GUI after a fresh install, so I decided to work from `main`. @mattersoflight we can diff before merging.

### DIFF
--- a/recOrder/napari.yaml
+++ b/recOrder/napari.yaml
@@ -1,18 +1,18 @@
-name: recOrder
+name: recOrder-napari
 schema_version: 0.1.0
 contributions:
   commands:
-  - id: recOrder.MainWidget
+  - id: recOrder-napari.MainWidget
     title: Create Main Widget
     python_name: recOrder.plugin.widget.napari_plugin_entry_point:MainWidget
-  - id: recOrder.get_reader
+  - id: recOrder-napari.get_reader
     title: Read ome-zarr files
     python_name: recOrder.io._reader:napari_get_reader
   readers:
-  - command: recOrder.get_reader
+  - command: recOrder-napari.get_reader
     accepts_directories: true
     filename_patterns: ['*.zarr']
   widgets:
-  - command: recOrder.MainWidget
+  - command: recOrder-napari.MainWidget
     display_name: Main Menu
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,7 +1,7 @@
 [metadata]
-name = recOrder
+name = recOrder-napari
 version = 1.0.0-beta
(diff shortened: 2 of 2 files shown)


mattersoflight:
Tested with a fresh environment and all of the above usage modes work.

ieivanov:
This renaming does break current installations of recOrder if changes are just pulled and napari/recOrder is launched. Reinstalling recOrder allows the plugin to start up, but I still get the following error:

napari.manifest -> 'recOrder:napari.yaml' could not be imported: Manifest name must match package name

talonchandler:
Thanks for testing @ieivanov. After more testing Shalin and I observed similar issues:

- this update breaks existing conda+pip installations of recOrder
- `pip install -e .` will restore most functionality with the warnings `napari.manifest -> 'recOrder:napari.yaml' could not be imported: Manifest name must match package name`. The issue seems to be related to two recOrder names in pip. `pip list` shows both `recOrder` and `recOrder-napari`. 
- All usage modes seem to work in a fresh conda environment. I'll update here if I find a set of commands that can restore functionality to an existing conda environment.

talonchandler:
#92 further expands on the symptoms @ieivanov and @Soorya19Pradeep were seeing. I think we can ignore this warning.

ieivanov:
I think the warning goes away if you delete the `recOrder-napari.egg-info` folder, but I'll have to check again

talonchandler:
Warning addressed in #95.

PR #94, opened 2022-04-16
UI/UX fixes

ieivanov:
Added rounding to logging functions, fixed enumeration of config groups to allow for groups other than "Channel"


talonchandler (approved):
Looks good to me!

PR #95, opened 2022-04-19
Build fixes; 'dev' mode

talonchandler:
Two build- and distribution-related fixes:

1. `include_package_data = True` moved from `metadata` to `options` so that non-python files are included in the wheels. This fix addresses manifest-related errors: issue #92, and discussion around #89. 
- Tested on MacOS and Windows

2. Moved **all** dependencies to `setup.cfg` and added a `dev` mode. 

    **Takeaway:** To install `dev` mode dependencies (like `wget`, `grip`, `pytest` etc) install with `pip install -e '.[dev]'`. 

- removed `test_requires` section of `setup.cfg` since it is deprecated ([or at least on it's way out](https://github.com/pypa/setuptools/issues/1684))
- removed package-by-package `pip install` commands from github actions
- GitHub actions are importing packages like `wget` correctly, but they are failing with HTTP 503---Zenodo seems to be down?


talonchandler:
Ready to merge from my perspective. I have successfully installed from `pip` and the `Plugins` menu, and I've updated the README with a "Easy Installation" and "Developer Installation" sections.

mattersoflight:
@talonchandler when I try to install recOrder-napari from the napari's plugin system, it ends up fetching the recOrder package from pypi. We may have to tweak the pypi name somewhere in the config.  Has this route of installation worked for you?

image

talonchandler:
Thanks for testing! I just tried a fresh install and I wasn't able to recreate this. 

Here's what I did so that we can narrow things down. On my local macOS I:

1) Created a fresh conda environment
2) `pip install napari[all]`
3) Clicked `Plugin > Install/Uninstall Plugins...`
4) Found `recOrder-napari` and clicked `Install`.
5) I didn't receive any feedback in the lower part of the window like you're seeing, but the recOrder-napari plugin appeared in the installed portion of the window and I was able to open the plugin. 

Maybe napari version? 

Screen Shot 2022-04-21 at 11 54 01 AM

talonchandler:
On my test 1 hour ago I used the top `filter...` box to find `recOrder-napari` then I clicked `Install`. 

I also just successfully installed `recOrder-napari` by typing `recOrder-napari` into the `install by name/url, or drop file...` box.

If I type `recOrder` into the `install by name/url, or drop file...` box then it tries (and fails) to install `recorder` (not the package we want). Is this what you did above?

talonchandler:
@mattersoflight following our in-person discussion I think this can be merged. I've now tested installation of recOrder through napari on Mac and Windows.

mattersoflight (approved):

PR #119, opened 2022-06-03
Pin pycromanager and MicroManager versions

talonchandler:
**Two changes:**
1. This addresses #115 by warning the user about MicroManager compatibility issues and explicitly recommends the 20210713 nightly build (the build we're currently using on all scopes in the lab). 
2. This downgrades and pins pycromanager to version 0.13.2, the latest version that is compatible with the MM 20210713 nightly build (ZMQ version 4.0.0). The ZMQ warnings disappear. 

Important related pycromanager discussion: https://github.com/micro-manager/pycro-manager/issues/260

Instead of the approach I'm suggesting here, @ieivanov suggested that we always pull the latest pycromanager version and manually update whichever MM install we're using by copy-pasting .jars. I tried this approach and it works, but I'm hesitant to use it because it requires managing the .jars and Henry recommended against it. 

Instead, I think we should do periodic updates like Henry suggests. We pick the latest MM version that works with our hardware and use the compatible pycromanager version. Every ~6 months, we can go through the effort of updating MM, testing our hardware, and updating pycromanager and recOrder. This means that each version of recOrder will be associated with a specific recommended version of MM and PM. 

I think one of Ivan's concerns is that we might want the latest features of MM and PM, but the MM device adaptors aren't compatible. If this is the case, then I think we should document and report the incompatibilities like Nico suggests. 

I will do more testing and wait for @ieivanov's comments (I hope I'm doing your concerns justice) before moving forward with this.

### DIFF
--- a/recOrder/plugin/widget/main_widget.py
+++ b/recOrder/plugin/widget/main_widget.py
@@ -1535,6 +1535,8 @@ class MainWidget(QWidget):
         -------
 
         """
+        RECOMMENDED_MM = '20210713'
+        ZMQ_TARGET_VERSION = '4.0.0'
         try:
             bridge = Bridge(convert_camel_case=False)
             self.mmc = bridge.get_core()
@@ -1543,8 +1545,11 @@ class MainWidget(QWidget):
             # Check MicroManager version compatibility.
             bridge._master_socket.send({"command": "connect", "debug": False}) # latest versions of pycromanager use '_main_socket'
             reply_json = bridge._master_socket.receive(timeout=500) # latest versions of pycromanager use '_main_socket'
-            if version.parse(reply_json['version']) < version.parse('4.0.0'):
-                print('WARNING: MicroManager version is incompatible with recOrder. Please upgrade to the latest tested MicroManager nightly build 20210713.')
+            zmq_mm_version = reply_json['version']
+            if zmq_mm_version != ZMQ_TARGET_VERSION:
+                upgrade_str = 'upgrade' if version.parse(zmq_mm_version) < version.parse(ZMQ_TARGET_VERSION) else 'downgrade'
+                print(("WARNING: This version of Micromanager has not been tested with recOrder.\n"
+                      f"Please {upgrade_str} to MicroManager nightly build {RECOMMENDED_MM}."))
 
             # Find and set calibration channel group
             calib_channels = ['State0', 'State1', 'State2', 'State3', 'State4']
--- a/recOrder/plugin/widget/main_widget.py
+++ b/recOrder/plugin/widget/main_widget.py
@@ -16,6 +16,7 @@ from recOrder.io.config_reader import ConfigReader, PROCESSING, PREPROCESSING, P
 from waveorder.io.reader import WaveorderReader
 from pathlib import Path, PurePath
 from napari import Viewer
+from packaging import version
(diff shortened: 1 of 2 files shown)


ieivanov:
I now agree that this is the most viable solution, thanks for testing @talonchandler. 

As a way of mitigating some of these issues, I would occasionally only upgrade specific device adapters - I would compile a new version and paste it in the MM 20210713 folder, or I would download the latest MM version and move the new device adapter to MM 20210713. This work as long as the MM Core version doesn't change between these nightly builds. We should keep a long of which device adapters we update in this manner, which can also guide when to update to a new MM nightly build.

recOrder/plugin/widget/main_widget.py
```diff
@@ -1539,6 +1540,12 @@ def connect_to_mm(self):
             self.mmc = bridge.get_core()
             self.mm = bridge.get_studio()
 
+            # Check MicroManager version compatibility.
+            bridge._master_socket.send({"command": "connect", "debug": False}) # latest versions of pycromanager use '_main_socket'
+            reply_json = bridge._master_socket.receive(timeout=500) # latest versions of pycromanager use '_main_socket'
+            if version.parse(reply_json['version']) < version.parse('4.0.0'):
+                print('WARNING: MicroManager version is incompatible with recOrder. Please upgrade to the latest tested MicroManager nightly build 20210713.')
+
```
ieivanov, line 1548: This checks if the MM ZMQ version is lower than 4.0.0. It's also possible to install a newer MM version and try to use it with recorder. We should add a check for ZMQ version grater than 4.0.0 and recommend downgrading to MM 20210713

ieivanov (approved):

recOrder/plugin/widget/main_widget.py
talonchandler: Good call. I've fixed this in the latest commits.

talonchandler:
Thanks Ivan!

I also tested my warning and recOrder behavior with a few versions of micromanager before and after 20210713. 

**MM versions before ~2021** will not establish a bridge. To help with these cases I added a warning message that hints at reasons why the bridge may not be connecting:
```
Could not establish pycromanager bridge.
Is micromanager open?
Is Tools > Options > Run server on port 4827 checked?
Are you using nightly build 20210713?"
```
**MM versions throughout ~2021** will "work" and prompt an MM upgrade. 
**MM versions newer than 20210713** will warn the user to downgrade, and different ZMQ versions give different behaviour: sometimes failing immediately, sometimes failing at acquisition time (like the serial spewing Soorya and I observed), and sometimes working okay.

talonchandler:
> We should keep a log of which device adapters we update in this manner, which can also guide when to update to a new MM nightly build.

All sounds good to me. My current plan for managing device drivers is to put them in the github release for each recOrder version. As a start, I've added the Meadowlark device drivers to [the 0.1.0 release](https://github.com/mehta-lab/recOrder/releases/download/0.1.0/recOrder-0.1.0-device-drivers.zip), and I'm happy to add other relevant drivers to this release if you think it makes sense.  

I've also linked to these drivers in the [new and improved installation guide](https://github.com/mehta-lab/recOrder/wiki/recOrder-Installation-and-MicroManager-Setup-Guide) that I'm using to support an installation at UW.

PR #127, opened 2022-06-30
Updates to LC calibration

ieivanov:
This PR refactors some of the LC calibration functions for easier control of the LCs by setting the LC retardance or LC voltage through MicroManager or changing the LC state based on voltages provided by a DAC.

New features:
* LC calibration by directly setting the LC voltage through MicroManager
* Extended CalibrationData class that enables Schnoor fit (https://doi.org/10.1364/AO.408383) to LC calibration data
* Consistent use of units - waves for retardance and volts for voltage, and added documentation to LC calibration functions 

Notebook showing measured intensity vs LC retardance using different LC control methods and retardance-to-voltage interpolation algorithms: https://github.com/czbiohub/compmicro-sandbox/blob/master/recOrder/LC_calibration/LC_Calibration_Intensity_vs_Retardance_Smoothness.ipynb

This PR works with v1.1.0 or later of the MeadowlarkLC device adapter as it uses the debugged LC voltage control method. This version of the device adapter (`mmgr_dal_MeadowlarkLcOpenSource.dll` and `mmgr_dal_MeadowlarkLC.csv`) is in `comp_micro/software/MM_Device_Adapters/MeadowlarkLC/DEVICE_INTERFACE_VERSION 70/FW 104/V1.1.0/`.


talonchandler:
Looking great!

> Should we expose the LC control mode (MM-Retardance, MM-Voltage, or DAC) and the retardance-to-voltage conversion method (interpolation interpolation vs Schnoor fit)?  If so, then where?

I think the "safest" option is to support both "interpolation" and "Schnoor fit" modes and add an extra GUI element so that the user can select the LC control mode and the conversion method.

But I'm also strongly in favor of simplifying if possible. I'm happy to commit to "Schnoor fit" mode if we test it properly. This will require more work/testing to build confidence before we merge, but it has the big advantage of simplifying the code and the user experience.

I think it's easy to keep adding options until we end up in spaghetti, so I'm in favor of simplifying by choosing "Schnoor fit" (and possibly dropping MM-Retardance mode so that the two modes are "MM" and "DAC") if we're willing to do the testing. 

> Where should we store the LC calibration data? Currently, in MM-Voltage and DAC control modes this version of the code uses calibration data stored in /recorder/calib/calibration_data.csv. This is not a good place to store it, as it's hard to access if recOrder is pip installed. A better place to store it would be the MM install folder, where mmgr_dal_MeadowlarkLC.csv is stored. I still have to check if MM-Retardance uses mmgr_dal_MeadowlarkLC.csv or the calibration data hardcoded in the device adapter.

I agree that putting the .csv into the MM folder makes sense, especially since moving the .csv into the MM folder is already part of the installation process.

talonchandler:
On thinking more, I think the bar for convincing ourselves that the Schnoor fit is working well might be a bit higher than the bar for the interpolation method. I think that a well-implemented Schnoor fit will be smoother and have less/no interpolation error, but I also think there's a risk that the Schnoor fit can have a much larger error if the fit doesn't work well in some regions.

To check the current fit performance, I just made a commit to this branch that adds an ad-hoc `plot_error` function. Feel free to revert/modify. I tested this with
```
$ python
>> from recOrder.calib import Calibration
>> cal_data = Calibration.CalibrationData('/Users/talon.chandler/recOrder/recOrder/calib/calibration_data.csv', interp_method='schnoor_fit')
>> cal_data.plot_error('error.png')
```
which gave:
error

This is slightly worse than Shnoor's figure 6 (max ~20 nm error compared to Schnoor's max ~10 nm error), but within our working voltages I think it's performing quite well. 

Is this good enough for what we're after? Strictly better than linear interpolation? (I'm not actually sure...but I think a retardance error might be a more informative filter than R^2 < 0.999).

ieivanov:
That plot collapses data from all 3 wavelengths, which is a bit confusing to look at. Here are plots of the calibration data at 546 nm, the Schnoor fit at that wavelength, and the associated fit error.

cal_data
error

As you see, the calibration data provided by Meadowlark is a bit discontinuous - jumps like the one near 2 V and 4 V are not quite physical. I'd say overall, the fit error is within +/- 10 nm. My check for `r_squared < 0.999` is mostly to ensure that the fit has worked well overall, rather than failing miserably.

Take a look at these data also: 

long_term_stability
long_term_stability_2

Thorlabs shows that the retardance vs voltage can vary quite a bit for their LC over time (and with temperature as we know), I'd expect the same for the Meadowlark LCs. 

The right thing to do here is not to rely on the calibration data as the "ground truth", but rather to calibrate the LCs each time we use them, as we do. To calibrate well, we need a continuous function that captures the general trend of the retardance vs voltage behavior of the LCs (as a function of wavelength also), which is what the Schnoor fit does. Overfitting the calibration data can lead to jumps in the retardance vs voltage as we have seen before. So we shouldn't look for absolute error relative to the calibration data, but rather for a smooth landscape of measured intensity vs LC retardance, which is on my TODO list.

ieivanov:
> I think the "safest" option is to support both "interpolation" and "Schnoor fit" modes and add an extra GUI element so that the user can select the LC control mode and the conversion method.

Ok, let's add a field for the retardance-to-voltage conversion method (maybe shorten the name a bit) to the GUI that we can test and keep until the end of the summer. I'd still argue for making `MM-Voltage` and `Schnoor fit` for making the default options in the drop-down menu. @talonchandler could you help me with that? I haven't used the QT Designer yet.

>I agree that putting the .csv into the MM folder makes sense, especially since moving the .csv into the MM folder is already part of the installation process.

I agree that's a good idea, but I think it may be a bit tricky to do. Can you help me figure out the MM install directory from within recOrder? Maybe we can get that thru pycromanager.

talonchandler:
Thanks again for being so patient on this PR. 

**Re: interpolation error.** Thanks for the extremely clear explanation! I'm 100% on board now...I completely agree that smoothness is more important than overfitting the noisy data. 

**Re: QT designer and adding a GUI interface.** I'm not deeply familiar with QT designer either (I've mostly been using it to help me understand the QT API), so I've made direct edits to `recOrder_ui.py` on top of the changes you made in this branch. 

This branch (and another branch I'm working on) include direct edits to the `recOrder_ui.py` file, so if we want to continue to support QT designer we will need to either recreate these changes in QT designer or move the changes to a different file. I'm personally fine with not supporting QT designer going forward, but maybe there are use cases that I'm not familiar with. FWIW I don't think Ahmet Can uses QT designer at all. 

**Re: MM paths:** I found the magic line we're looking for: `mmc.getDeviceAdapterSearchPaths().get(0)` and committed it to this branch. 

**Re: testing status of this branch:** I've done very basic tests on my machine, but I haven't tested with full calibrations. Would you like me to do some testing on automaton?

ieivanov:
Notebook showing measured intensity vs LC retardance using different LC control methods and retardance-to-voltage interpolation algorithms: https://github.com/czbiohub/compmicro-sandbox/blob/master/recOrder/LC_calibration/LC_Calibration_Intensity_vs_Retardance_Smoothness.ipynb

ieivanov:
@mattersoflight @talonchandler @Soorya19Pradeep this PR is ready for review and merge.

To use the code, make sure you update the Meadowlark device adapter to the latest version, which can be found in `comp_micro/software/MM_Device_Adapters/MeadowlarkLC/DEVICE_INTERFACE_VERSION 70/FW 104/V1.1.0/` (see info at top of PR)

I tried to think of better names for the LC control modes, but couldn't think of something that's both short and descriptive. Instead, I added documentation at the init for the QLIPPCalibration class and also mouse hover pop-ups in the GUI

Screen Shot 2022-08-04 at 11 14 18 AM

recOrder/calib/Calibration.py
```diff
@@ -771,12 +878,14 @@ def write_metadata(self, notes=None, microscope_params=None):
 
         metadata = {'Summary': {
                         'Timestamp': str(datetime.now()),
-                        'recOrder-napari version': str(pkg_resources.require("recOrder-napari")[0]),
-                        'waveorder version': str(pkg_resources.require("waveorder")[0])},
+                        'recOrder-napari version': version('recOrder-napari'),
```
talonchandler, line 881: Much better, thank you!

recOrder/calib/Calibration.py
```diff
@@ -907,39 +1020,193 @@ def capture_bg(self, n_avg, directory):
...
+        y = ydata.flatten()
+        y_hat = y - p.fun
+        slope, intercept, r_value, *_ = linregress(y, y_hat)
+        r_squared = r_value**2
+        if r_squared < 0.999:
+            warnings.warn(f'Schnoor fit has R2 value of {r_squared:.5f}, fit may not have worked well.')
+
+        return p.x
+
+    def interpolate_data(self, raw_data, calib_wavelengths):
```
talonchandler, line 1198: `calib_wavelengths` is unused. I'm guessing that you you were thinking of rewriting in a more generalizable way (instead of a hardcoded small set of wavelengths)? I'm fine with skipping work and just deleting this option since `schnoor_fit` is performed well.

recOrder/plugin/qtdesigner/recOrder_ui.py
```diff
@@ -262,6 +262,7 @@ def setupUi(self, Form):
         self.cb_calib_mode.setObjectName("cb_calib_mode")
         self.cb_calib_mode.addItem("")
         self.cb_calib_mode.addItem("")
+        self.cb_calib_mode.addItem("")
```
talonchandler, line 265: My commit moves all changes in this file to `main_wiget.py` to allow QtCreator to be used in the future.

recOrder/plugin/widget/main_widget.py
```diff
@@ -1551,19 +1552,33 @@ def connect_to_mm(self):
                 config_list = []
                 for j in range(configs.size()):
                     config_list.append(configs.get(j))
-                if np.all([ch in config_list for ch in calib_channels]):
-                    config_group_found = True
-                    self.config_group = group
+                if np.all([np.any([ch in config for config in config_list]) for ch in self.calib_channels]):
```
talonchandler, line 1555: Would you mind adding a comment here on the intended behaviour? This is a tricky section.

I think you're trying to catch cases where there's a typo in one of the 'stateX' strings when you're configuring micromanager?

recOrder/plugin/widget/main_widget.py
```diff
@@ -1551,19 +1552,33 @@ def connect_to_mm(self):
...
+                msg = f'No config group contains channels {self.calib_channels}. ' \
                       'Please refer to the recOrder wiki on how to set up the config properly.'
                 self.ui.cb_config_group.setStyleSheet("border: 1px solid rgb(200,0,0);")
                 raise KeyError(msg)
+
+
+            # set LC control mode
+            _devices = self.mmc.getLoadedDevices()
+            loaded_devices = [_devices.get(i) for i in range(_devices.size())]
+            if lc_device_name in loaded_devices:
```
talonchandler, line 1573: Very nice default behavior!

talonchandler (approved):
Excellent work! I enjoyed going through in detail...I definitely learned a few things.

I have added commits that 
- Populate the GUI calibration modes and tooltips from your (extremely nice) docstring. This is to reduce duplication and allows us to use QtCreator going forward. 
- Capitalized LC_DEVICE_NAME everywhere to indicate that it's being treated as a global constant.

recOrder/calib/Calibration.py
ieivanov: Yeah, I was hoping to make the linear fit a bit more generic, but I figures it's not worth the time at this point. I think I kept it in just to mimic the signature of the `fit_data` method. Maybe we can keep it in and just leave a comment in the code for now?

talonchandler: Works for me!

recOrder/plugin/widget/main_widget.py
```diff
@@ -53,6 +56,20 @@ def __init__(self, napari_viewer: Viewer):
...
+
+        # Populate calibration modes from docstring
+        cal_docs = NumpyDocString(Calibration.QLIPP_Calibration.__init__.__doc__)
+        mode_docs = ' '.join(cal_docs['Parameters'][3].desc).split('* ')[1:]
+        for i, mode_doc in enumerate(mode_docs):
+            mode_name, mode_tooltip = mode_doc.split(': ')
+            wrapped_tooltip = '\n'.join(textwrap.wrap(mode_tooltip, width=70))
+            self.ui.cb_calib_mode.addItem(mode_name)
+            self.ui.cb_calib_mode.setItemData(i, wrapped_tooltip, Qt.ToolTipRole)
+
```
ieivanov, line 73: Oh, that's a cool trick! Thanks!!

recOrder/plugin/widget/main_widget.py
ieivanov: Just added some documentation above this section: 

https://github.com/mehta-lab/recOrder/blob/081e59ae5f6fa2077505b9980074c8b6a2530e2c/recOrder/plugin/widget/main_widget.py#L1583-L1589

ieivanov:
I think this is ready to merge?

talonchandler:
Ready to merge from my perspective!

PR #139, opened 2022-08-05
Correct 'global' and 'local_fit' background corrections

### LINKED ISSUE #350: Rename 'global' and 'local' background correction to 'measured' and 'estimated'
Split from mehta-lab/recOrder#86:

Soorya: "The names 'Global' and 'local' background correction in the background reconstruction section of recorder are counterintuitive. The names 'measured' background for corrections using background images, and 'estimated' background for fitted background is more intuitive."

After discussion with @mattersoflight and @ieivanov we agreed that this does not need to be a breaking change. We can change recOrder's GUI interface in the short term while leaving the config files unchanged.

### LINKED ISSUE #367: [BUG] Global background correction doesn't always work
@ieivanov and I had a debugging session after lunch and realized that the "Global" background correction wasn't working as expecting. During this morning's sessions @mattersoflight and I only applied "Local" bkg corrections, so we missed this---good call Ivan.

I've just finished a long debugging session, and I haven't been able to track down the bug. I'm observing some surprising behavior that I'll continue to try and track down. There are some "brittle" workarounds documented below. 

To test, I've been attempting to correct the background of a background image. I expect reasonably uniform retardance and fairly random orientations. My "DOES NOT WORK" steps generate a result indistinguishable from not applying any background correction. My "WORKS" steps give results that are better than applying no background correction, but they are still very imperfect: uniform retardance (~7 [units]) across the FOV and orientations pointing predominantly towards pi/2.

@ieivanov if you have time tomorrow morning I'd love to have another session to see if any of these results look familiar. Thanks!

----

1. DOES NOT WORK (this is what we did earlier, @ieivanov)
- start recOrder
- **calibrate LCs**
- **acquire bkg images**
- attempt background-corrected birefringence acquisition. 

2. DOES NOT WORK
- start recOrder
- **load calibration from file**
- **load background from file in acquisition tab**
- attempt background-corrected birefringence acquisition. 

3. WORKS (EFFECTIVE WORKAROUND)
- start recOrder
- **load background from file in acquisition tab**
- **load calibration file**
- attempt background-corrected birefringence acquisition. 

4. WORKS
- start recOrder
- **load calibration from file**
- **acquire bkg images**
- attempt background-corrected birefringence acquisition.

talonchandler:
`recOrder`'s background correction was not behaving in the way that I expected---see attached figure. I tested recOrder's GUI using this [test-config.yml](https://github.com/mehta-lab/recOrder/files/9264320/test-config.txt) on the zenodo data.

'local' and 'global' corrections were giving identical results while 'none' seemed to work (see first row). 

bkg-correction

Three simultaneous fixes were required:

1. To fix the background correction in `recOrder` alone, I needed to use a kludge to conform to `waveorder`'s `Polscope_bg_correction` function. Specifically, `Polscope_bg_correction` requires and applies an array-based background correction no matter what background correction option you choose. To work around this for the `local` option, I supplied an "identity" background correction. 

AFAIK most users have been avoiding `Polscope_bg_correction` by applying their corrections "by hand" in scripts. I think myself, @Soorya19Pradeep, and @JohannaRahm are doing this based on scripts I've seen. I think this is a strong argument for either rewriting `Polscope_bg_correction` or refactoring this functionality, but here I'm working under the constraint of a fixed `waveorder` dependency. 

2. The 'global' background correction was averaging over the FOV even when an ROI was not supplied. 

**Details (skippable):** This bug was caused by JSON converting python tuples to lists. A tuple was compared to a list which resulted in "False" when the result should have been "True". Specifically, the `utils.bg_read` function averaged over an ROI area when it shouldn't, which in turn made recOrder's BG correction incorrect (or at least different from expected). This means that both 'local' and 'global' corrections were behaving incorrectly. 

3. Background corrections weren't propagating to 3D stacks correctly for 'local_fit'.

The `local_fit` logic in `waveorder` depends on `self.N_defocus` (L1181 `waveorder_reconstructor.py`). My first choice would be to improve `waveorder` since the background correction code should have close to zero dependencies on the constructor object...but I'm only applying fixes to `recOrder` here so I supplied more options to the `waveorder` constructor so that `self.N_defocus` is correct. 

---

The result is matching my expectations (see second row of figure), though I'd appreciate input from @ieivanov and @mattersoflight. I will test on hummingbird before merging. 

I am fine with these fixes for 0.2.0, but I would like to consider cleaning up `waveorder` for 1.0.0 because this was a challenging bug to fix and I expect the background correction code to continue to cause headaches.

### DIFF
--- a/recOrder/pipelines/qlipp_pipeline.py
+++ b/recOrder/pipelines/qlipp_pipeline.py
@@ -79,6 +79,7 @@ class QLIPP(PipelineInterface):
                                                           wavelength_nm=self.config.wavelength,
                                                           swing=self.calib_meta.Swing,
                                                           calibration_scheme=self.calib_scheme,
+                                                          n_slices=self.data.slices,
                                                           pad_z=self.config.pad_z,
                                                           bg_correction=self.config.background_correction,
                                                           mode=self.mode,
--- a/recOrder/pipelines/qlipp_pipeline.py
+++ b/recOrder/pipelines/qlipp_pipeline.py
@@ -72,6 +72,12 @@ class QLIPP(PipelineInterface):
         self.data_shape = (self.t, len(self.output_channels), self.slices, self.img_dim[0], self.img_dim[1])
         self.chunk_size = (1, 1, 1, self.data_shape[-2], self.data_shape[-1])
 
+        # Convert 'local_fit+' to 'local_fit' for waveorder
+        if self.config.background_correction == 'local_fit+':
+            wo_background_correction = 'local_fit'
+        else:
+            wo_background_correction = self.config.background_correction
+
         # Initialize Reconstructor
         if self.no_phase:
             self.reconstructor = initialize_reconstructor(pipeline='birefringence',
@@ -81,7 +87,7 @@ class QLIPP(PipelineInterface):
                                                           calibration_scheme=self.calib_scheme,
                                                           n_slices=self.data.slices,
                                                           pad_z=self.config.pad_z,
-                                                          bg_correction=self.config.background_correction,
+                                                          bg_correction=wo_background_correction,
                                                           mode=self.mode,
                                                           use_gpu=self.config.use_gpu,
                                                           gpu_id=self.config.gpu_id)
@@ -100,22 +106,22 @@ class QLIPP(PipelineInterface):
                                                           z_step_um=self.data.z_step_size,
                                                           pad_z=self.config.pad_z,
                                                           pixel_size_um=self.config.pixel_size,
-                                                          bg_correction=self.config.background_correction,
+                                                          bg_correction=wo_background_correction,
                                                           mode=self.mode,
                                                           use_gpu=self.config.use_gpu,
                                                           gpu_id=self.config.gpu_id)
 
-        # Compute BG stokes if necessary
-        if self.config.background_correction == 'global':
-            print("Loading bg from "+self.bg_path)
+        # Prepare background corrections for waveorder
+        if self.config.background_correction in ['global', 'local_fit+']:
             bg_data = load_bg(self.bg_path, self.img_dim[0], self.img_dim[1], self.bg_roi)
             self.bg_stokes = self.reconstructor.Stokes_recon(bg_data)
             self.bg_stokes = self.reconstructor.Stokes_transform(self.bg_stokes)
         elif self.config.background_correction == 'local_fit':
             self.bg_stokes = np.zeros((5, self.img_dim[0], self.img_dim[1]))
-            self.bg_stokes[0, ...] = 1 # Set background to "identity" Stokes parameters. waveorder's 'local_fit' applies this identity correction.
+            self.bg_stokes[0, ...] = 1  # Set background to "identity" Stokes parameters.
         else:
             self.bg_stokes = None
+
 
     def _check_output_channels(self, output_channels):
         self.no_birefringence = True
--- a/recOrder/pipelines/qlipp_pipeline.py
+++ b/recOrder/pipelines/qlipp_pipeline.py
@@ -105,11 +105,14 @@ class QLIPP(PipelineInterface):
                                                           gpu_id=self.config.gpu_id)
 
         # Compute BG stokes if necessary
-        if self.config.background_correction != 'None':
+        if self.config.background_correction == 'global':
+            print("Loading bg from "+self.bg_path)
             bg_data = load_bg(self.bg_path, self.img_dim[0], self.img_dim[1], self.bg_roi)
             self.bg_stokes = self.reconstructor.Stokes_recon(bg_data)
             self.bg_stokes = self.reconstructor.Stokes_transform(self.bg_stokes)
-
+        elif self.config.background_correction == 'local_fit':
+            self.bg_stokes = np.zeros((5, self.img_dim[0], self.img_dim[1]))
+            self.bg_stokes[0, ...] = 1 # Set background to "identity" Stokes parameters. waveorder's 'local_fit' applies this identity correction.
         else:
             self.bg_stokes = None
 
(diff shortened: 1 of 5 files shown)


recOrder/pipelines/qlipp_pipeline.py
```diff
@@ -105,11 +106,14 @@ def __init__(self, config: ConfigReader, data: WaveorderReader, writer: Waveorde
...
+            print("Loading bg from "+self.bg_path)
             bg_data = load_bg(self.bg_path, self.img_dim[0], self.img_dim[1], self.bg_roi)
             self.bg_stokes = self.reconstructor.Stokes_recon(bg_data)
             self.bg_stokes = self.reconstructor.Stokes_transform(self.bg_stokes)
-
+        elif self.config.background_correction == 'local_fit':
+            self.bg_stokes = np.zeros((5, self.img_dim[0], self.img_dim[1]))
+            self.bg_stokes[0, ...] = 1 # Set background to "identity" Stokes parameters. waveorder's 'local_fit' applies this identity correction.
         else:
             self.bg_stokes = None
```
ieivanov, line 119: Thanks for debugging this @talonchandler!

In the most common mode of reconstruction, we first apply the measured background (aka 'global' correction) and if needed we further fit and subtract a surface from this once already background corrected data. This is what I call "local" background correction, but we've seen that these names are poor. In practice it's a combination of "global" correction plus local fit.

The "global" background correction should simply subtract the background images, as is implemented here.

In this implementation, "local" background correction only subtracts the fitted 2d surface. I think it's better to rename this method and apply what I described at the top.

There are cases where we cannon acquire a background image and perform "global" background correction. In this case, we would only do surface fitting. You've implemented this here as providing unity background images.

So I think we need a total of four background correction methods:
* "none" - no background correction
* "measured" - use only background images
* "measured+" - use background images and surface fit to remove any residual background
* "estimated" - only use surface fit

talonchandler: Aha! This context really helps me see why the waveorder code looks the way it does. I 100% agree with needing four options (either through a single four-option menu, two two-option menus, or two checkboxes). 

I was originally thinking of putting the "renaming of options" into it's own PR, but with this comment I think now I'll do my renaming here.

talonchandler:
I've implemented a first-pass at what Ivan suggested with the following user exposed names and tooltips:

None: No background correction
Measured: Correct with a measured background from file.
Estimated: Correct with a background estimated from the raw image.'
Measured + Estimated: Correct with a measured background from file then correct further with an estimated background.
        
Screen Shot 2022-08-05 at 5 11 44 PM

Here are the results on the test data. Although the "Estimated" and "Measured + Estimated" look similar, they are subtly different...the "Estimated" correction just seems to dominate for this example. Does this match your experience for samples like this, Ivan? I don't have much experience with the fitting to judge whether or not this is working well (although I'm tracing the code and seeing it hit the "right places"). 

g439

I will test on hummingbird on Monday, implement bug fixes, then I'll mark as ready for review. In the meantime, comments are welcome on my naming choices (@Soorya19Pradeep) and the results from the test data.

talonchandler:
I forgot to mention that these new names are only in the GUI because I need to maintain backwards compatibility. I left the old names unchanged in the config files and added one new option `local_fit+` which corresponds to the new "Measured + Estimated" background correction GUI option.

ieivanov:
> Although the "Estimated" and "Measured + Estimated" look similar, they are subtly different...the "Estimated" correction just seems to dominate for this example. Does this match your experience for samples like this, Ivan?

Yes, this looks right to me. On a sample like this where you have large correlation in S1 and S2 over the field of view the right thing to do is to apply measured background correction, rather than a local fit. Local fit works well where structures in the sample are isotropically distributed on the scale of the field of view (think of cultured cells). When S1 and S2 have large correlations (imagine a FOV with sarcomeres all running in the same direction) then the local fit will try to eliminate those, which will be wrong. In the Kazansky dataset here, the 2D surface fit is affected by the structures of the target and you see larger background than just using the measured background. Note that your perception of random orientation may be skewed here by what colors we perceive as bright.

ieivanov:
Suggestion for rewording the tooltips. Method names are still up for discussion, I think.

None: No background correction
Measured: Correct sample images with a background image acquired at an empty field of view, loaded from Background Path.
Estimated: Estimate sample background by fitting a 2D surface to the sample images. Works well when structures are isotropically distributed on the scale of the field of view.
Measured + Estimated: Apply Measured background correction and then Estimated background correction. Use to remove any residual background between the two datasets.

talonchandler:
Thanks for iterating on the tooltips, Ivan. My latest iteration of names and tooltips (with my minor changes...all open for discussion) are:

**None:** No background correction.
**Measured:** Correct sample images with a background image acquired at an empty field of view, loaded from "Background Path".
**Estimated:** Estimate sample background by fitting a 2D surface to the sample images. Works well when structures are spatially distributed across the field of view and a clear background is unavailable.',
 **Measured + Estimated** Apply "Measured" background correction and then "Estimated" background correction. Use to remove residual background from "Measured"-corrected datasets.

Here are the background corrections applied to a background image with histograms:

g3245

No major surprises from my perspective. 

I also had a session on hummingbird where I tested the background correction behavior in "Online" and "Offline" mode...no surprises. 

I'm changing from draft to PR now, since I think this is ready for review+merge. All comments appreciated!

mattersoflight:
Thanks @talonchandler for these fixes and @ieivanov for the review.

The refined terminology (measured, estimated, measured + estimated) reads well.  A suggestion for a tooltip for 
Measured + Estimated correction: `Apply "Measured" background correction and then "Estimated" background correction. Use to remove residual background after the sample retardance is corrected with measured background.`

It looks like this PR takes care of issues #104 and #136. I've noted recent ideas for improving background correction on an older issue (#4).  This PR has also has useful notes on how waveOrder's background correction can be improved.
  
I think the PR is ready to merge.

mattersoflight (approved):

talonchandler:
Thanks for reading + rephrasing, Shalin, and thanks for bringing up #4. 

I'll let the tests run and wait for Ivan's last word before merging.

ieivanov:
Sorry, I'm behind on this. Looks good to me too, thanks for merging.

PR #140, opened 2022-08-05
Restructure tests

### LINKED ISSUE #88: Tests require larges files as fixtures
I've been trying to run and debug the main branch's tests with `pytest`, and I think a large fraction of the testing time is spent downloading the test files from zenodo. 

Surprisingly, the GitHub actions tests run quickly. Does GitHub keep a copy somewhere? Can I use a local copy for the tests?

Possible solutions (in my rough order of preference) are:

**1) Use smaller test files.** This would be best for running fast tests, but maybe some of the tests depend on large files? My intuition says this is the best option since the testing should be quick, easy, and as local as possible. This could be an easy-ish fix by using a small FOV within the same data and adding that data (a few MBs) to the repo. 

**2) Continue to use the same files,** but report the download progress to the tester. I thought the tests were broken until I dug through the code and noticed that the downloads were the bottleneck. Should be a fairly easy addition if we choose this. 

**3) Be patient with the tests.** Definitely possible, but my machine is currently taking 5 mins to run a single test (~10 mins for all of them) because it downloads the files each time. 

If I'm doing something dumb, then I'll make the testing workflow clearer in the docs https://github.com/mehta-lab/recOrder/wiki#development-tools

ieivanov:
fixes #88

### DIFF
--- a/recOrder/io/zarr_converter.py
+++ b/recOrder/io/zarr_converter.py
@@ -6,9 +6,8 @@ from waveorder.io.writer import WaveorderWriter
 from waveorder.io.reader import WaveorderReader
 from recOrder.preproc.pre_processing import get_autocontrast_limits
 from recOrder.io.utils import create_grid_from_coordinates
-import glob
+import copy
 import json
-import warnings
 
 
 class ZarrConverter:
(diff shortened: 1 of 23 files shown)


ieivanov:
Notes: in zarr_converter.py change 
self.data_type = data_type -> self.reader.data_type
self.dtype = self.reader.reader.dtype -> self.reader.dtype

ieivanov:
@talonchandler this PR is ready for review

recOrder/io/zarr_converter.py
```diff
@@ -19,31 +18,41 @@ class ZarrConverter:
     for tiled acquisitions)
     """
 
-    def __init__(self, input, output, data_type=None, replace_position_names=False, format_hcs=False):
+    def __init__(self, input_dir, output_dir, data_type=None, replace_position_names=False, format_hcs=False):
```
talonchandler, line 21: Much clearer name choice. 

I think the user-facing command-line options should match this choice, but that will break backwards compatibility, so I'll mark it as a 1.0.0 change.

recOrder/io/zarr_converter.py
```diff
@@ -19,31 +18,41 @@ class ZarrConverter:
...
+        replace_position_names: bool
+        format_hcs: bool
+        """
 
         # Add Initial Checks
-        if len(glob.glob(os.path.join(input, '*.tif'))) == 0:
-            raise ValueError('Specific input contains no .tif files, please specify a valid input directory')
-        if not output.endswith('.zarr'):
+        # if len(glob.glob(os.path.join(input, '*.tif'))) == 0:
+        #     raise ValueError('Specific input contains no .tif files, please specify a valid input directory')
```
talonchandler, line 38: Still worth keeping? I've seen this error as a user when I wasn't sure if I should give a file or a directory (your appending `_dir` to helps this problem), and I think I'd prefer this message over a cryptic IO failure if I accidentally supplied a folder without any tifs.

talonchandler (approved):

recOrder/io/zarr_converter.py
talonchandler: See #147

talonchandler:
Ready to go from my perspective! Looks great, @ieivanov.

recOrder/io/zarr_converter.py
ieivanov: I think this check should be implemented in the reader, rather than the converter, the reader should decide if the data format is valid. For example, pycromanager datasets are structures as `input_dir/Full Resolution/*.tif`. The waveorder `MicromanagerOmeTiffReader` implements this check here: https://github.com/mehta-lab/waveorder/blob/cbdba709594649a3eac50ea7c9adc8ec71c0d0cf/waveorder/io/multipagetiff.py#L23-L25

PR #146, opened 2022-08-11
Convert GUI-acquired data to `zarr` before reconstruction 

### LINKED ISSUE #365: [FEATURE] Write raw data to disk before reconstruction starts
**Problem**
At AQLM, we observed that raw data is written to the disk after reconstruction ends, which can lead to loss of data. 
**Proposed solution**
The default behavior should be to convert raw data into a zarr store and write to the disk before reconstruction starts. This choice aligns with our upcoming refactor to separate acquisition and reconstruction in two separate tabs.

talonchandler:
The recOrder GUI's current behavior is to 
(1) acquire MM data,
(2) perform and save the reconstruction, then
(3) convert the raw MM data to zarr. 

This PR swaps the order of (2) and (3) so that the conversion to zarr happens before the reconstruction. If the reconstruction fails (for OOM or other reasons) this PR keeps the converted data so that it's ready to go for another attempt. 

I've tested locally on all four acquisition buttons.

### DIFF
--- a/recOrder/plugin/workers/acquisition_workers.py
+++ b/recOrder/plugin/workers/acquisition_workers.py
@@ -145,6 +145,9 @@ class BFAcquisitionWorker(WorkerBase):
         stack = acquire_from_settings(self.calib_window.mm, settings, grab_images=True)
         self._check_abort()
 
+        # Cleanup acquisition by closing window, converting to zarr, and deleting temp directory
+        self._cleanup_acq()
+
         # Reconstruct snapped images
         self.n_slices = stack.shape[2]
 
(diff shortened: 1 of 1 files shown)


ieivanov (approved):

ieivanov:
Looks good to me.

Data reconstruction can sometimes take a while, and it can look like the GUI is not doing anything (or worse, has crashed) if we close the acquisition window and then nothing happens. I know that there are messages that pop up on the terminal, but these may not always be visible. It will be helpful to get a message on the GUI that say "Reconstructing data..." or something of this soft to provide a bit of user feedback. Feel free to leave this till after the GUI refactor.

talonchandler:
@ieivanov good point...I agree. I've opened a new issue #153 for 1.1.0 so we don't forget. Thanks!

Issue #149, opened 2022-08-23
[BUG]Fails to run on Windows


nicost:
**Summary and expected behavior**

Follows installation instructions to the letter on Windows 10:  Installed Anaconda, 
```
conda create -n recOrder python
conda activate recOrder
pip install "napari[all]"
conda create -n recOrder python
conda activate recOrder
git clone https://github.com/mehta-lab/recOrder.git
cd recOrder
pip install -e ".[dev]"
```
Running napari and selecting the recOrder plugin results in:

`RuntimeError: Failed to import command at 'recOrder.plugin.widget.napari_plugin_entry_point:MainWidget': No module named 'importlib_metadata'`

**Code for reproduction (using python script or command line interface)**
```
See above
```

**Screenshots or steps for reproduction (using napari GUI)**

**Expected behavior**
I would hope that the plugin would start without an error

**Environment:**

Operating system: Windows 10 Pro
Python version: 3.10.4
Python environment (command line, IDE, Jupyter notebook, etc):
MicroManager/pycromanager version:

talonchandler:
Hi Nico! Thanks for reporting. 

I just reproduced the error and the problem is a difference between `importlib_metadata` and `importlib.metadata` (the first is an external package and the second's a builtin) that slipped by us. I'll open a PR right now.  

If you're okay with using our latest release, you can install with:
```
conda deactivate
conda env remove -n recOrder
conda create -n recOrder python
conda activate recOrder
pip install "napari[all]"
pip install recOrder-napari 
napari -w recOrder-napari
```
which I just tested on my Windows 10 machine. 

Let me know how it goes...thanks for your patience.

PR #150, opened 2022-08-23
Require importlib-metadata

talonchandler:
#149 was introduced by #127, and this fixes the issue. 

This error snuck by my review because I assumed `importlib_metadata` was a builtin. My bad. I'm surprised this snuck by our tests...I will look closer tomorrow morning and update the tests to catch. 

I tested a successful install by hand. Ready to merge from my perspective.


talonchandler:
I've added a simple smoke test to prevent this from happening again. Tests failed on the branch from `main` and now pass after merging into this branch. 

Ready for review + merge.

Issue #108, opened 2022-05-04
[BUG] Reconstruction fails in the middle of processing due to insufficient RAM


mattersoflight:
At AQLM, we had a computer with 64GB RAM. This computer could not handle retardance & phase reconstruction  of 2kx2kx50 stack. It is optimal to allocate memory for all variables required for reconstruction before reconstruction starts. This issue may need to be moved to waveOrder if allocation is done by methods there.

talonchandler:
Reading between the lines of #107 and #108, I'm guessing that you requested a big acquisition+reconstruction then received OOM during the reconstruction you lost everything---definitely a major issue.

I think the full solution will involve:
1) Solving #107 so that data is always saved to disk after acquisition. After this is fixed is solved we can request expensive recons without worrying about this type of catastrophe.
2) Like you suggest: preallocating the arrays so that waveorder fails early and gracefully if you ask too much. I expect this will involve moving the preallocation calls to the `waveorder.waveorder_microscopy` constructor within `try`/`except` blocks.

talonchandler:
3. Recreating and testing this bug might be a bit tricky, but [this discussion](https://stackoverflow.com/questions/41105733/limit-ram-usage-to-python-program) has some useful snippets for artificially limiting the amount of RAM available to python.

talonchandler:
I'm flagging this issue for discussion because I think it will take more effort and give less reward than I initially thought. 
We may've already reached a satisfactory solution by saving to disk before reconstruction starts (issue #107 solved by PR #146). 

Here are two solutions we've discussed:

**Solution 1: Preallocate all of the arrays so that a large reconstruction fails early.**

This solution will work, and I have only minor concerns about the effort required to preallocate the correct arrays in `waveorder` since there are many paths through the `waveorder_microscopy` class. 

However, my **major concern is that preallocating the arrays will make our memory performance worse**. By tying up all of our memory early in a reconstruction we will have less free memory to work with. We can partially sidestep this problem by manually freeing up memory when we're finished with it, but this is additional work and only a bandage since we'll be using a *lot* of memory at the start of the recon. 

**Solution 2: Find empirical heuristics for when memory usage will be too large, and warn the user if they make a request that's larger than the RAM available.**

[...]

PR #157, opened 2022-08-27
Print RAM warning message in online and offline modes

talonchandler:
This PR prints the following message to users who request a reconstruction (via online or offline modes) when <32 GB of RAM is available:

"WARNING: recOrder reconstructions often require more than the 16 GB of RAM that this computer is equipped with. We recommend starting with reconstructions of small volumes ~1000 x 1000 x 10 and working up to larger volumes while monitoring your RAM usage with Task Manager or htop."

When >32 GB is available (e.g. 64 GB), it prints:

"64 GB of RAM is available."

I'm happy to take wording edits or completely overhaul this PR given our discussion in #108. 

Note: `psutil` is a dependency of one of our dependencies, so it is already installed in `recOrder` environments.

### DIFF
--- a/recOrder/io/utils.py
+++ b/recOrder/io/utils.py
@@ -1,5 +1,7 @@
 import glob
 import os
+import psutil
+import textwrap
 import tifffile as tiff
 import numpy as np
 from waveorder.waveorder_reconstructor import fluorescence_microscopy, waveorder_microscopy
@@ -165,3 +167,16 @@ def get_unimodal_threshold(input_image):
             max_dist = per_dist
     assert best_threshold > -np.inf, 'Error in unimodal thresholding'
     return best_threshold
+
+def ram_message():
+    BYTES_PER_GB = 2**30
+    gb_available = psutil.virtual_memory().total / BYTES_PER_GB
+    if gb_available < 32:
+        return ' \n'.join(textwrap.wrap(
+               f'\033[91mWARNING:\033[0m recOrder reconstructions often require more than the {gb_available:.1f} ' \
+               f'GB of RAM that this computer is equipped with. We recommend starting with reconstructions of small ' \
+               f'volumes ~1000 x 1000 x 10 and working up to larger volumes while monitoring your RAM usage with '
+               f'Task Manager or htop.',
+        ))
+    else:
+        return f'{gb_available:.1f} GB of RAM is available.'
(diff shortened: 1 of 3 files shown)


talonchandler:
This is ready to merge from my perspective. 

@mattersoflight does this adequately address #108? I don't think there's much need to use a more complete/complex solution since we're saving to disk before reconstruction.

recOrder/io/utils.py
```diff
@@ -165,3 +167,16 @@ def get_unimodal_threshold(input_image):
             max_dist = per_dist
     assert best_threshold > -np.inf, 'Error in unimodal thresholding'
     return best_threshold
+
+def ram_message():
+    BYTES_PER_GB = 2**30
+    gb_available = psutil.virtual_memory().total / BYTES_PER_GB
+    if gb_available < 32:
+        return ' \n'.join(textwrap.wrap(
+               f'\033[91mWARNING:\033[0m recOrder reconstructions often require more than the {gb_available:.1f} ' \
```
ziw-liu, line 176: I'm not sure if we ever use `logging` in the CLI. If we do maybe it would make sense to not manually format the message?

ziw-liu: Also ANSI code support on Windows seems to be version-specific.

talonchandler: Good call on `print` vs. `logging`. I just changed the online mode `print`s to `logging.info`. 

Re: ANSI code support: I tested that it works in powershell, but you might be right with cmd.exe. I think the worst that will happen is that you'll see "\033[91mWARNING:\033[0m", which isn't the end of the world, so I'm not too concerned.

ziw-liu (approved):

recOrder/io/utils.py
ziw-liu: We will see something like `[Warning] WARNING ...` in online mode, but that's only cosmetic.

ziw-liu: I did not think it through, we are now actually seeing:
```
10:21:26 INFO WARNING: recOrder reconstructions often require more than the
16.0 GB of RAM that this computer is equipped with. We recommend
starting with reconstructions of small volumes ~1000 x 1000 x 10 and
working up to larger volumes while monitoring your RAM usage with Task
Manager or htop.
```
And this message will only be displayed in console/log, without flashing a warning in the GUI.
Maybe we should have it do `show_warning(ram_message())` instead.

ziw-liu: Addressed in #174

PR #133, opened 2022-07-29
Improve help messages for `recOrder`'s CLI

talonchandler:
This PR makes the CLI help statements much more accessible. The relevant help message will print if you:
- type `recOrder` or `recorder` without any arguments (to catch new users with typos)
- append `-h` or `--help` or any unrecognized argument
- type `recOrder.convert` or `recOrder.reconstruct`
- use any combination of the above three (e.g. `recorder.reconstruct -h` will work)

This PR also simplifies the entry points by putting them in a single file, renaming the functions, and removing the ArgParse class structures.

PR #158, opened 2022-08-27
CLI regression fix and end-to-end CLI test

talonchandler:
I accidentally introduced a regression that broke the CLI in #133 (I failed to move `import os` into the new file). 

This PR fixes the regression and adds a test that runs the CLI in a background process and checks if it finishes successfully. This type of end-to-end test will give us more confidence that the CLI can run at all. Even though we were unit testing the functions underlying the CLI calls, the CLI was unusable and this bug made it past the tests.

### DIFF
--- a/recOrder/scripts/cli.py
+++ b/recOrder/scripts/cli.py
@@ -1,5 +1,6 @@
 import click
 import sys
+import os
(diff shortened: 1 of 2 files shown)


talonchandler:
The fix is working and the tests are passing on all platforms. Ready for review + merge.

PR #159, opened 2022-08-27
Don't delete test data?

talonchandler:
@ieivanov I owe an apology for deleting these lines a few months ago without much announcement, then you were likely confused and had to reintroduce them. 

I'm suggesting deleting them again because I would like to run new/modified tests locally without waiting for the test data to download. 

I'm suggesting that we keep the test data, but you may have a better solution. Is there a reason to delete them? I don't think GitHub actions complains? Is there a different workflow that you use to quickly iterate and rerun the tests without waiting for a complete download?

### DIFF
--- a/recOrder/tests/conftest.py
+++ b/recOrder/tests/conftest.py
@@ -26,13 +26,6 @@ def setup_test_data():
 
     yield test_data
 
-    try:
-        # remove temp folder
-        shutil.rmtree(temp_folder)
-    except OSError as e:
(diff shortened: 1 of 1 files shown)


ieivanov (approved):
No worries, that's fine with me. Should we do the same for `waveorder` as well?

talonchandler:
Good call. I just made a matched PR in waveorder: https://github.com/mehta-lab/waveorder/pull/83

Issue #120, opened 2022-06-15
[FEATURE] Refactor recOrder calibration


ieivanov:
Refactor `recOrder ` calibration module such that
* it is independent of LC voltage control device - TriggerScope or NIDaq
* it is independent of image acquisition software - MicroManager, pycromanager, or coPylot
* it can be imported and used by external libraries

cc: @AhmetCanSolak

AhmetCanSolak:
hello @ieivanov ,

Thank you for starting this discussion here. I quickly went over the the `recOrder.calib` module and I believe it can become independent from acquisition software or voltage control devices. I felt like in the existing code many places can become more independent. An example: https://github.com/mehta-lab/recOrder/blob/b5328d2ed4b8c2deb596988a11954e76c163adae/recOrder/calib/Optimization.py#L13-L25

This small method of the BrentOptimizer can take `current_lca` and `current_lcb` values as arguments then this method wouldn't need to  know about `self.calib.mmc`. We can apply similar thinking on other parts too and achieve what you desire.

Regarding your third point:
```
it can be imported and used by external libraries
```
I like to propose hosting new calibration module in coPylot at root of the package(something like `copylot.calib`) instead of having new calibration code as a dependency and importing it(in a way better to have less packages to maintain IMO). We can still make the calibration code lighter so  it can be important on other places from copylot when needed easily yet code can live in copylot repository so we  know where to find and  follow changes on it easily. Looking forward to hearing what do you think about this @ieivanov ?

ieivanov:
> This small method of the BrentOptimizer can take current_lca and current_lcb values as arguments then this method wouldn't need to know about self.calib.mmc. We can apply similar thinking on other parts too and achieve what you desire.

Totally agree. In this example also, the `0.01` and `1.6` limits are hardcoded, it would be helpful if these are defined as constants (`LC_RETARDANCE_MIN` and `LC_RETARDANCE_MAX` at the top of the class.

More generally, I think the `recOrder.calib` module should be responsible for choosing the next LC state in the optimization procedure given the current LC state and measured image intensity. Actually changing the LC state and acquiring images at those new settings should happen outside of `recOrder.calib`. Does that make sense? Do you agree @AhmetCanSolak @mattersoflight and @talonchandler? This will allow us to make LC optimization, data acquisition, and hardware control independent. We'll have to see how feasible that is.

> I like to propose hosting new calibration module in coPylot at root of the package(something like copylot.calib)

[...]

PR #160, opened 2022-08-29
Update shutter behavior and LF acquisition channel order

### LINKED ISSUE #111: [FEATURE] acquire data in channel-first order and keep the shutter open during interactive acquisition
**Problem**
Currently recOrder acquires data in the z-first order and toggles the shutter between each frame (if auto-shutter is chosen). The z-first order increases the chances of misalignment between polarization channels when there is drift in stage or when the live sample is being imaged. Opening/closing the shutter can slow down the acquisition of dynamics. 

**Proposed solution**
It is better to set the MDA in channel-first mode and keep the shutter open when label-free channels are acquired. 

**Additional context**
Noted during AQLM 2022

### LINKED ISSUE #142: Calibration fails when a shutter isn't available
Yuming from the Elicieri group had trouble calibrating his LCs because of a poor black level estimate. We traced this problem to a his lack of a shutter. We found a manual workaround: turn off the light source during black level estimation, then turn it back on for the rest of the calibration. 

I think a more complete solution will include:
- a warning to the user if a shutter isn't available
- a GUI element that lets the user input a black level by hand. This should be a pretty stable value, so I expect this to be a good solution.

### LINKED ISSUE #346: Calibration fails when a shutter isn't available
Yuming from the Elicieri group had trouble calibrating his LCs because of a poor black level estimate. We traced this problem to a his lack of a shutter. We found a manual workaround: turn off the light source during black level estimation, then turn it back on for the rest of the calibration. 

I think a more complete solution will include:
- a warning to the user if a shutter isn't available
- a GUI element that lets the user input a black level by hand. This should be a pretty stable value, so I expect this to be a good solution.

### LINKED ISSUE #361: [FEATURE] acquire data in channel-first order and keep the shutter open during interactive acquisition
**Problem**
Currently recOrder acquires data in the z-first order and toggles the shutter between each frame (if auto-shutter is chosen). The z-first order increases the chances of misalignment between polarization channels when there is drift in stage or when the live sample is being imaged. Opening/closing the shutter can slow down the acquisition of dynamics. 

**Proposed solution**
It is better to set the MDA in channel-first mode and keep the shutter open when label-free channels are acquired. 

**Additional context**
Noted during AQLM 2022

ieivanov:
This PR fixes mehta-lab/waveorder#361. It also starts to address ~#110~ and mehta-lab/waveorder#346 

To fully fix mehta-lab/waveorder#346 we also need to:
* ask the user to manually close and open the shutter if an auto shutter is not available
* provide the option to manually enter a black level  - either in the GUI or the CLI
@talonchandler could you please help me with those? 

~Towards addressing mehta-lab/waveorder#362 I've added a check to ensure that all LF channel exposures are the same. If they are not, all exposures are set to the exposure for State0. That's not a complete solution, but will prevent acquiring corrupted data.~


recOrder/calib/Calibration.py
```diff
@@ -639,13 +644,41 @@ def opt_I135(self, lca_bound, lcb_bound):
...
         logging.debug("--------done--------")
 
-    def calc_blacklevel(self):
+    def close_shutter(self):
+        self._auto_shutter_state = self.mmc.getAutoShutter()
+        self._shutter_state = self.mmc.getShutterOpen()
+
+        if not self.shutter_device:
+            # ask user to close shutter
+            pass
```
ieivanov, line 653: @talonchandler would appreciate your help here

recOrder/calib/Calibration.py
```diff
@@ -639,13 +644,41 @@ def opt_I135(self, lca_bound, lcb_bound):
...
+        else:
+            self.mmc.setAutoShutter(False)
+            self.mmc.setShutterOpen(False)
+
+    def open_shutter(self):
+        if not self.shutter_device:
+            # ask user to open shutter
+            pass
```
ieivanov, line 661: and here

recOrder/calib/Calibration.py
```diff
@@ -639,13 +644,41 @@ def opt_I135(self, lca_bound, lcb_bound):
...
+
+        Returns
+        -------
 
-        self.mmc.setAutoShutter(False)
-        self.mmc.setShutterOpen(False)
+        """
+        if not self.shutter_device:
+            # ask user to open shutter
+            pass
```
ieivanov, line 675: and here too :)

recOrder/plugin/widget/main_widget.py
```diff
@@ -1799,21 +1799,25 @@ def handle_fluor_reconstructor_update(self, value):
...
+        #         current_json['Microscope Parameters'][key] = meta[key]
+        #     else:
+        #         current_json['Microscope Parameters'][key] = None
+        #
+        # with open(self.last_calib_meta_file, 'w') as file:
+        #     json.dump(current_json, file, indent=1)
 
-        with open(self.last_calib_meta_file, 'w') as file:
-            json.dump(current_json, file, indent=1)
+        pass
```
ieivanov, line 1820: I've realized that the acquisition works try to do a weird update of the calibration metadata. I think the intent is to update the microscope parameters (such as NA or magnification) in the metadata file. However, we currently (at least in most of my use cases) don't enter such microscope parameters during LC calibration, and they are also not written at that time - `write_metadata` is typically called without the optional `microscope_params` argument. So this function will have no effect, and in fact it throws an error since we changed the `Microscope Parameters` key to `Microscope parameters`. 

The only time we currently write these microscope parameters is during background capture: https://github.com/mehta-lab/recOrder/blob/1318f2d50d9cb9057822bdb2e6016ec79ac05f3a/recOrder/plugin/workers/calibration_workers.py#L321

I think it's good to save some of these parameters (such as magnification, n_objective_media, objective_NA, condenser_NA, swing, pixel_size) in the calibration metadata. Parameters such as pad_z or bg_correction_option belong to the dataset metadata, not the calibration metadata.

I welcome help and input on this issue

recOrder/plugin/workers/acquisition_workers.py
```diff
@@ -878,6 +878,22 @@ def work(self):
...
             self._check_abort()
 
+            # Check that all LF channels have the same exposure settings
+            channel_exposures = []
+            for i in range(len(settings['channels'])):
+                channel_exposures.append(settings['channels'][i]['exposure'])
+
+            channel_exposures = np.array(channel_exposures)
+            if not np.all(channel_exposures == channel_exposures[0]):
+                # TODO: warn user that not all channels exposures were the same
```
ieivanov, line 888: @talonchandler could you help me with a warning message here? I think a warning on the CLI only is enough, right?

talonchandler:
This is excellent, Ivan, thanks so much. I'll definitely be able to take this over the finish line. 

I really like your "warning + default to State0" solution for #110---good call. 

In the interests of onboarding Ziwen, I'm thinking of testing and completing the fixes for #110, #111, and #142 on three different branches off of this branch, merge them into this single PR, then merge this into `main`. Sound okay? Thanks for giving us a major headstart!

recOrder/plugin/widget/main_widget.py
talonchandler: Great, thanks for the context here. I'll think more about what parameters should be saved alongside the calibration metadata. I'll probably break it off into a separate issue/PR since there's a bit of work to do here.

recOrder/plugin/workers/acquisition_workers.py
talonchandler: I think a warning at the command line is enough. I also need to make sure that the metadata is written correctly, but that should be easy.

talonchandler:
I've moved the exposure check to a different PR #162 since the implementation needs work, and this PR will benefit from narrowed scope.

talonchandler:
This PR is on the larger side. It includes:

**Features**
- keeping the shutter open during multi-channel and multi-slice acquisitions
- options for users to calibrate when a shutter isn't available
  - if a shutter isn't available, a warning appears prompting an input on the command line
  - the user can enter a number that will be used as a black level, or close the shutter manually and let recOrder estimate the black level
- slice-first acquisition instead of channel-first acquisition

**Non-features**
- To allow for testing of import statements, I have made `napari[all]` a requirement (and updated the docs accordingly).
- Dropped support for `python 3.7` because `napari` has also dropped support. 

@ziw-liu much of this depends on hardware, but I'd still appreciate a quick second set of eyes on the code before merging. Thanks!

.github/workflows/test_and_deploy.yml
```diff
@@ -21,7 +21,7 @@ jobs:
     strategy:
       matrix:
         platform: [ubuntu-latest, windows-latest, macos-latest]
-        python-version: [3.7, 3.8, 3.9]
+        python-version: [3.8, 3.9]
```
ziw-liu, line 24: The `python-version` specified here is not consistent with `pytests.yml`, namely lacking Python 3.10. Are we supporting for 3.10?

recOrder/calib/Calibration.py
```diff
@@ -145,6 +146,11 @@ def __init__(self, mmc, mm, group='Channel', lc_control_mode='MM-Retardance', in
         self.directory = None
         self.inst_mat = None
 
+        # Shutter
+        self.shutter_device = self.mmc.getShutterDevice()
```
ziw-liu, line 150: What is the expected return value if no shutter is available to the MMCore?

ziw-liu: From `QLIPP_Calibration.open_shutter` it seems like an empty string is expected. Need testing with hardware.

recOrder/calib/Calibration.py
```diff
@@ -639,13 +645,41 @@ def opt_I135(self, lca_bound, lcb_bound):
...
+            self.mmc.setAutoShutter(self._auto_shutter_state)
+            self.mmc.setShutterOpen(self._shutter_state)
+
+    def close_shutter_and_calc_blacklevel(self):
+        self._auto_shutter_state = self.mmc.getAutoShutter()
+        self._shutter_state = self.mmc.getShutterOpen()
+
+        if self.shutter_device == '':  # no shutter
+            show_warning('No shutter found. Please follow the command-line instructions...')
+            in_string = input('recOrder could not find an automatic shutter configured through MicroManager.\n'
```
ziw-liu, line 674: How about:

```python
shutter_warning_msg = """
recOrder could not find an automatic shutter configured through Micro-Manager.
>>> If you would like manually enter the black level, enter an integer or float and press <Enter>
>>> If you would like to estimate the black level, please close the shutter and press <Enter> 
"""

in_string = input(shutter_warning_msg)
```

recOrder/calib/Calibration.py
```diff
@@ -655,15 +689,8 @@ def calc_blacklevel(self):
...
-
-        self.mmc.setAutoShutter(auto_shutter)
-
-        if not auto_shutter:
-            self.mmc.setShutterOpen(shutter)
-
         self.I_Black = blacklevel
-
-        return blacklevel
+        return
```
ziw-liu, line 693: No need for explicit `return` here.

recOrder/plugin/workers/acquisition_workers.py
```diff
@@ -872,9 +873,16 @@ def work(self):
...
                                              save_dir=self.snap_dir,
-                                             prefix=self.prefix)
+                                             prefix=self.prefix,
+                                             keep_shutter_open_channels=True,
+                                             keep_shutter_open_slices=True)
 
             self._check_abort()
+
+            # set acquisition order to channel-first
+            settings['slicesFirst'] = False
```
ziw-liu, line 883: Sorry that my [#162](https://github.com/mehta-lab/recOrder/pull/162) code caused this conflict, but you will need `self.settings` here.

recOrder/calib/Calibration.py
```diff
@@ -639,13 +645,41 @@ def opt_I135(self, lca_bound, lcb_bound):
...
-        auto_shutter = self.mmc.getAutoShutter()
-        shutter = self.mmc.getShutterOpen()
+        Returns
+        -------
 
-        self.mmc.setAutoShutter(False)
-        self.mmc.setShutterOpen(False)
+        """
+        if self.shutter_device == '': # no shutter
+            input('Please reset the shutter to its original state and press <Enter>')
```
ziw-liu, line 663: Do we want to direct the user back to the GUI here? (e.g. `logging.info("This is the end of the commandline instructions.")`)

tox.ini
```diff
@@ -1,12 +1,11 @@
 # Modified from from cookiecutter-napari-plugin
 # For more information about tox, see https://tox.readthedocs.io/en/latest/
 [tox]
-envlist = py{37, 38,39}-{linux,macos,windows}
+envlist = py{38,39}-{linux,macos,windows}
```
ziw-liu, line 4: Also no Python 3.10 here.

.github/workflows/test_and_deploy.yml
talonchandler: Good catch! Fixed.

tox.ini
talonchandler: Fixed!

recOrder/calib/Calibration.py
talonchandler: Much nicer! Thanks.

recOrder/calib/Calibration.py
talonchandler: Nice!

recOrder/calib/Calibration.py
talonchandler: Great idea!

recOrder/plugin/workers/acquisition_workers.py
talonchandler: Not your fault! Great catch.

ziw-liu (approved):

ieivanov:
Quick note: I see that you've removed `run_calibration`, `run_5state_calibration`, and `run_4state_calibration` from `calib.Calibration`. These function used to let users run LC calibration outside of the `recOrder` widget (e.g. in a script). I think that's OK, as this is not a commonly used mode of `recOrder` right now, and we don't test and debug it.

talonchandler:
Thanks for the context @ieivanov. 

I'm totally in favor of script-based calibration if it's useful. For the sake of maintenance/updates, it will be great to have a single function that the GUI/scripts/others can call. I think we can revisit these design choices when we address #120.

PR #162, opened 2022-08-31
Check for equal exposures across channels

### LINKED ISSUE #110: recOrder uses MDA exposure settings which can create mostly unusable polarization data
When MM makes a multi-channel acquisition it lets you set different exposures for each channel. This feature is necessary for general MM use, but can lead to surprising/confusing behavior during polarization acquisitions via recOrder. 

Problems arise when the user:
1) Opens MM with the default exposure (say, 10 ms). 
2) Clicks "Acquire Birefringence" from recOrder.
3) (Naively) changes the exposure (say to 5 ms) from the MM main window. 
4) Clicks "Acquire Birefringence" and get weird results because the acquisition uses (5, 10, 10, 10) ms exposures. 

This behaviour accidentally bit @Soorya19Pradeep yesterday, and I can foresee it biting many new users in the future. I've also had to annoyingly reenter exposure times into the MDA window during sessions by myself and with @ieivanov.

I am considering two possible solutions:

1) Set the exposure from recOrder via a line edit box. I think this is the most modular design and will lead to the most predictable results. 
2) Warn the user if they request a polarization MDA with different exposure times.

-----

Extra context: in `recOrder/acq/acq_functions.py` lines 63-64 are:
```
#todo: think about how to deal with missing exposure
exposure = app.getChannelExposureTime(channel_group, chan, 10) # sets exposure to 10 if not found
```
I think we should only write to the MDA from recOrder, instead of reading and writing.

talonchandler:
This PR fixes #110, and was split from #160 (started by @ieivanov). 

This change checks for equality of exposure time across channels. If the exposures are not equal it uses the exposure associated with State0 and TODO: warns the user. 

This initial implementation only applies to 3D birefringence acquisitions. TODO: the final implementation should apply to all birefringence acquisitions.

### DIFF
--- a/recOrder/plugin/workers/acquisition_workers.py
+++ b/recOrder/plugin/workers/acquisition_workers.py
@@ -919,7 +919,8 @@ class PolarizationAcquisitionWorker(WorkerBase):
         channel_exposures = np.array(channel_exposures)
         if not np.all(channel_exposures == channel_exposures[0]):
             # warn user that not all channels exposures were the same
-            warn_exposure_msg = 'The exposure times of each State are not equal!\nAcquiring with the exposure of State0...'
+            warn_exposure_msg = f'The MDA exposure times are not equal!\n' \
+                                f'Acquiring with State0 exposure = {channel_exposures[0]} ms.'
             show_warning(warn_exposure_msg)
 
             # setting all channel exposures to the exposure of State0
--- a/recOrder/plugin/workers/acquisition_workers.py
+++ b/recOrder/plugin/workers/acquisition_workers.py
@@ -924,10 +924,6 @@ class PolarizationAcquisitionWorker(WorkerBase):
         """
         Acquire images.
 
-        Parameters
-        ----------
-        settings:       (json) JSON dictionary conforming to MM SequenceSettings
-
         Returns
         -------
         stack:          (nd-array) Dimensions are (C, Z, Y, X). Z=1 for 2D acquisition.
--- a/recOrder/plugin/workers/acquisition_workers.py
+++ b/recOrder/plugin/workers/acquisition_workers.py
@@ -912,7 +912,7 @@ class PolarizationAcquisitionWorker(WorkerBase):
         logging.debug('Verifying exposure times...')
         # parse exposure times
         channel_exposures = []
-        for _, channel in enumerate(self.settings['channels']):
+        for channel in self.settings['channels']:
             channel_exposures.append(channel['exposure'])
 
         channel_exposures = np.array(channel_exposures)
@@ -921,8 +921,8 @@ class PolarizationAcquisitionWorker(WorkerBase):
             logging.warning('The exposure times of each State are not equal!\nAcquiring with the exposure of State0...')
 
             # setting all channel exposures to the exposure of State0
-            for i in range(len(self.settings['channels']) - 1):
-                self.settings['channels'][i + 1]['exposure'] = self.settings['channels'][0]['exposure']
+            for i in range(len(self.settings['channels'])):
+                self.settings['channels'][i]['exposure'] = self.settings['channels'][0]['exposure']
         
         self._check_abort()
 
--- a/recOrder/plugin/workers/acquisition_workers.py
+++ b/recOrder/plugin/workers/acquisition_workers.py
@@ -851,23 +851,21 @@ class PolarizationAcquisitionWorker(WorkerBase):
             logging.debug('Acquiring 2D stack')
 
             # Generate MDA Settings
-            settings = generate_acq_settings(self.calib_window.mm,
+            self.settings = generate_acq_settings(self.calib_window.mm,
                                              channel_group=self.channel_group,
                                              channels=channels,
                                              save_dir=self.snap_dir,
                                              prefix=self.prefix)
             self._check_abort()
-
-            # Acquire from MDA settings uses MM MDA GUI
-            # Returns (1, 4/5, 1, Y, X) array
-            stack = acquire_from_settings(self.calib_window.mm, settings, grab_images=True)
+            # acquire images
+            stack = self._acquire()
 
         # Acquire 3D stack
         else:
             logging.debug('Acquiring 3D stack')
 
             # Generate MDA Settings
-            settings = generate_acq_settings(self.calib_window.mm,
+            self.settings = generate_acq_settings(self.calib_window.mm,
                                              channel_group=self.channel_group,
                                              channels=channels,
                                              zstart=self.calib_window.z_start,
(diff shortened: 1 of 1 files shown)


talonchandler:
Hi @ziw-liu , I've peeled this PR off from Ivan's larger PR, and I think this will be a good first issue for you to tackle. 

Reading #110 and #160 will be good places to start, and we can can work together on Monday to move this forward. 

You'll be able to test almost all of this without microscope hardware, so I think this issue will be a good forcing function to help you get a local test environment setup.

ziw-liu:
### Changes with [last commit](https://github.com/mehta-lab/recOrder/pull/162/commits/e18467518888df70bb1eea8f00348846623b0ef4)
1. Lifted the exposure check to the `PolarizationAcquisitionWorker._check_exposure` method.
2. Acquisition `settings` is now a property of `PolarizationAcquisitionWorker`.
3. Lifted the checking and acquiring actions to the `PolarizationAcquisitionWorker._acquire` method.

recOrder/plugin/workers/acquisition_workers.py
```diff
@@ -906,6 +901,54 @@ def work(self):
         self.phase_image_emitter.emit(phase)
         self.meta_emitter.emit(meta)
 
+    def _check_exposure(self) -> None:
```
talonchandler, line 904: I like pulling this out as it's own function...looks good and makes sense. Nice docs, too.

ziw-liu:
### TODO
Flash warning in the Napari GUI.
Offer GUI interface for setting exposure?

recOrder/plugin/workers/acquisition_workers.py
```diff
@@ -906,6 +901,54 @@ def work(self):
...
+            # warn user that not all channels exposures were the same
+            logging.warning('The exposure times of each State are not equal!\nAcquiring with the exposure of State0...')
+
+            # setting all channel exposures to the exposure of State0
+            for i in range(len(self.settings['channels']) - 1):
+                self.settings['channels'][i + 1]['exposure'] = self.settings['channels'][0]['exposure']
+        
+        self._check_abort()
+
+    def _acquire(self) -> np.ndarray:
```
talonchandler, line 929: I'm unsure about pulling this out as it's own function since it's only 3 lines, it only gets called once, and it those three lines do more than just "acquire". I'm also unsure about this refactor since it doesn't match the other classes in this file.

recOrder/plugin/workers/acquisition_workers.py
```diff
@@ -906,6 +901,54 @@ def work(self):
...
+        Check that all LF channels have the same exposure settings. If not, use the State0 exposure.
+
+        Parameters
+        ----------
+        settings:       (json) JSON dictionary conforming to MM SequenceSettings
+        """
+        logging.debug('Verifying exposure times...')
+        # parse exposure times
+        channel_exposures = []
+        for _, channel in enumerate(self.settings['channels']):
```
talonchandler, line 915: Simplify to `for channel in self.settings['channels']:`?

recOrder/plugin/workers/acquisition_workers.py
```diff
@@ -906,6 +901,54 @@ def work(self):
...
+            channel_exposures.append(channel['exposure'])
+
+        channel_exposures = np.array(channel_exposures)
+        if not np.all(channel_exposures == channel_exposures[0]):
+            # warn user that not all channels exposures were the same
+            logging.warning('The exposure times of each State are not equal!\nAcquiring with the exposure of State0...')
```
talonchandler, line 921: I think you mentioned you're looking into a napari-level warning? If so I think this is great.

recOrder/plugin/workers/acquisition_workers.py
```diff
@@ -906,6 +901,54 @@ def work(self):
...
+
+            # setting all channel exposures to the exposure of State0
+            for i in range(len(self.settings['channels']) - 1):
+                self.settings['channels'][i + 1]['exposure'] = self.settings['channels'][0]['exposure']
```
talonchandler, line 925: Consider changing to
```
for i in range(len(self.settings['channels'])):
    self.settings['channels'][i]['exposure'] = self.settings['channels'][0]['exposure']
```
so that a reader doesn't have to think through all of the off-by-one possibilities.

talonchandler (commented):
Looking great!

talonchandler:
> Offer GUI interface for setting exposure?

To avoid in-the-weeds GUI stuff you could use a very simple `input` statement like this:
https://github.com/mehta-lab/recOrder/pull/160/commits/9792dc4433d1a95ffce3707a92da619ee4a531a4

Completely up to you.

recOrder/plugin/workers/acquisition_workers.py
ziw-liu: Yes that makes sense

recOrder/plugin/workers/acquisition_workers.py
ziw-liu: It gets called two in two places (2D and 3D) so I think reducing repetitive code might save some future errors.

talonchandler: Roger that...makes sense. Thanks!

recOrder/plugin/workers/acquisition_workers.py
ziw-liu: That was the original code. My thought was that Python probably does not have compile-time-optiization that takes out `self = self`. But shouldn't matter anyway since this is cheap.

ziw-liu:
> To avoid in-the-weeds GUI stuff you could use a very simple `input` statement like this: [9792dc4](https://github.com/mehta-lab/recOrder/commit/9792dc4433d1a95ffce3707a92da619ee4a531a4)

What I was concerned about was that the user may not pay attention to the command lines when operating in Napari. So I think a visual notice in the GUI is necessary, and we should then choose between notifying them to set in MM and offering an input box somewhere in the Napari window.

recOrder/plugin/workers/acquisition_workers.py
ziw-liu: Will look into that with future commits

recOrder/plugin/workers/acquisition_workers.py
ziw-liu: Changed it back to the simpler way

talonchandler:
Yeah I see what you mean. I'm leaning towards just a warning message...no need to provide extra input options if MM already provides the interface IMO.

ziw-liu:
On second thought: If we make sure that the user sees a notification on the Napari window, then is it even safer to throw an error and exit the worker without proceeding with assumptions about the user's intentions?

talonchandler:
I lean towards a warning instead of an error in this case. Presumably the user the user wants some sort of acquisition to happen if they've hit "Acquire"...and I think a sensible default (use State0's exposure) together with a warning is a good solution.

ziw-liu:
When the `recOder-napari` changes the MDA exposure times, the new exposures are updated in the Micro-manager MDA window, but in not updated where the `recOrder-napari` reads the acquisition settings from. A second acquisition worker after the first warning (and setting equal acquisition times) will still detect inconsistent exposures.

ziw-liu:
If we do not intend to fix the current MDA updating behavior, do we want to throw an error and abort the acquisition worker instead of sending a warning?

ziw-liu:
Branch that only sends a warning: [exposure-check-warning-only](https://github.com/mehta-lab/recOrder/tree/exposure-check-warning-only)

talonchandler:
>> throw an error and abort the acquisition worker instead of sending a warning

I think this is a good plan. Thanks Ziwen.  

Extra context for Ivan/Shalin in case they read this later:
- setting MDA exposures from pycromanager created bugs that would be very confusing to the user (e.g. all of the exposures would be equal, but we would still get unequal exposure warnings because MM/MDA is caching the exposures somewhere). 
- since we're planning to remove acquisitions from recOrder in 1.0.0, we decided to simplify our solution and give an error if the user requests an MDA with non-equal exposures.

talonchandler:
Looks great! Just tested locally, and everything is working as I would expect. 

I made a small update the error message. 

@ziw-liu I approve...you can let the tests pass then merge.

ziw-liu (approved):
Now behaving as expected. Resolving the issue.

PR #163, opened 2022-08-31
Don't update calibration metadata

talonchandler:
This commit is cherry picked from #160 to narrow it's scope. 

Context from that thread @ieivanov:
Screen Shot 2022-08-31 at 3 18 10 PM

### DIFF
--- a/recOrder/plugin/widget/main_widget.py
+++ b/recOrder/plugin/widget/main_widget.py
@@ -1799,21 +1799,25 @@ class MainWidget(QWidget):
 
     @pyqtSlot(dict)
     def handle_meta_update(self, meta):
-        if self.last_calib_meta_file is None:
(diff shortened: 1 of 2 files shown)


talonchandler:
I agree with @ieivanov that rewriting and/or overwriting any `calibration_metadata` file at acquisition time is a surprising behavior. 

As Ivan suggested, I have left the `handle_meta_update` function commented so that we only write the `calibration_metadata` file at calibration time. 

I have also reduced the number of microscope parameters that are written when you acquire a background.
`pad_z`, `bg_correction_option`, and `swing` are no longer written. 
`n_objective_media`, `objective_NA`, `condenser_NA`, `magnification`, and `pixel_size` are still written.

ziw-liu (approved):

PR #165, opened 2022-09-07
OSX_arm64 dev install and Markdown linting

ziw-liu:
Added documentation to address compatibility issues during developer installation on Apple Silicon (OS X `arm_64`) platforms.
There are also minor improvements made to the Markdown source code formatting.


docs/software-installation-guide.md
```diff
@@ -3,45 +3,89 @@
...
+    pip install -e ".[dev]"
+    ```
+
+5. To acquire data via `Micro-Manager`, follow  [microscope installation guide](./microscope-installation-guide.md).
 
 **Optional GPU**: `recOrder` supports NVIDIA GPU computation with the `cupy` package. Follow [these instructions](https://github.com/cupy/cupy) to install `cupy` and check its installation with ```import cupy```. To enable gpu processing, set ```use_gpu: True``` in the config files.
 
+### Mac OS X on Apple Silicon (arm_64)
+
+[Napari](napari.org) uses PyQt5 to draw its GUI, which is not currently compatible with `pip`. We have tested the following emulated development environment on Mac OS 12.5. (It is available natively  through [conda-forge](https://github.com/conda-forge/miniforge) if the user does not intend to edit the `recOrder-napari` source code.)
```
talonchandler, line 62: - "PyQt5...is not currently compatible with `pip`" --- is this true? I think my current `recOrder` environment is running a `pip install`ed PyQt5?
- "It is available natively" - does "It" refer to PyQt5 here?
- extra space between natively and through

docs/software-installation-guide.md
```diff
@@ -3,45 +3,89 @@
...
+
+1. Install and enable [Rosetta](https://support.apple.com/en-us/HT211861) for your terminal of choice. To verify, enter command `arch` in shell and it should print 'i386' instead of 'arm64'; or  check if the terminal process is labeled as 'Type: Intel' in the Activity Monitor.
```
talonchandler, line 64: I'm still a bit confused about what is happening in your environment. When I type `arch` I get `arm64` and everything runs smoothly, so I don't think this is necessary. FWIW in Activity Monitor I see 'Type: Apple' for my terminal.

docs/software-installation-guide.md
```diff
@@ -3,45 +3,89 @@
...
+
+2. (Optional) For connection with `Micro-Manager`, install Java (JDK 8 `x86_64`) and make sure it is properly linked for your system Java wrappers. For example, OpenJDK 8 may be installed as follows if you have a `x86_64` installation of [Homebrew](https://brew.sh/):
```
talonchandler, line 66: Hmmm....are you using an `x86_64` version of `brew`?

docs/software-installation-guide.md
```diff
@@ -3,45 +3,89 @@
...
+
+    ```sh
+    arch -x86_64 /usr/local/bin/brew install openjdk@8
+    ```
+
+3. Create a conda environment:
+
+    ```sh
+    CONDA_SUBDIR=osx-64 conda create -n [environment] python=3.9  # create a new environment
```
talonchandler, line 75: FWIW I haven't had to use `CONDA_SUBDIR` on my M1. If it's possible to find a simpler workaround that would be great...there must be a different between our environments that's causing difficulties with your install.

docs/software-installation-guide.md
```diff
@@ -3,45 +3,89 @@
...
+    ```sh
+    cd recOrder
+    pip install -e ".[dev]"
+    ```
+
+5. To acquire data via `Micro-Manager`, follow  [microscope installation guide](./microscope-installation-guide.md).
 
 **Optional GPU**: `recOrder` supports NVIDIA GPU computation with the `cupy` package. Follow [these instructions](https://github.com/cupy/cupy) to install `cupy` and check its installation with ```import cupy```. To enable gpu processing, set ```use_gpu: True``` in the config files.
 
+### Mac OS X on Apple Silicon (arm_64)
```
talonchandler, line 60: Should `arm_64` -> `i386`? The rest of these instructions seem like they're for `i386`, but maybe I'm a bit confused.

docs/software-installation-guide.md
```diff
@@ -3,45 +3,89 @@
...
+
+```sh
 recOrder.help
 ```
 
-## Developer installation:
+## Developer installation
 
-Install [`git`](https://git-scm.com/book/en/v2/Getting-Started-Installing-Git) and `conda` (either [anaconda](https://www.anaconda.com/products/distribution) or [miniconda](https://docs.conda.io/en/latest/miniconda.html)).
+### Windows (x86_64)
```
talonchandler, line 32: and M1 `arm64` since these instructions have been working for me?

talonchandler (commented):
Thanks for documenting your install, Ziwen!

I'm still not clear what created such large differences between our systems: I haven't had to use `arch` or `CONDA_SUBDIR` during my installations, so I'm wondering if there's a baseline Rosetta setting that different between our systems?

If these instructions are for M1 `i386`, then I approve and we can merge. If they're for M1 `arm_64` then I'm still a bit confused and we might be able to simplify.

docs/software-installation-guide.md
ziw-liu: I meant PyQt5 is not available natively with `pip`. You are right that the expression is not very clear here.

docs/software-installation-guide.md
ziw-liu: I am. `brew` refuses to install certain `x86_64` packages if it was installed as a `arm64` version (`brew` on different architecture have different default installation directories).

docs/software-installation-guide.md
ziw-liu: This was to denote the system's 'native' architecture, but should also work if we delete the '(arm_64)' here.

docs/software-installation-guide.md
ziw-liu: Not exactly the behavior in my tests. I think we need more testing here to see what is the 'default' workflow (which may have changed due to variations in shipped OS setting, etc.)

docs/software-installation-guide.md
ziw-liu: Conda (or at least Conda-forge distribution) should in theory automatically detect the 'correct' build of packages depending on the architecture of the system. I explicitly set `$CONDA_SUBDIR` here to ensure consistency in package versions and hence perhaps better reproducibility, but I do suspect some of them are redundant.

docs/software-installation-guide.md
ziw-liu: Verified that terminal/shell architectures do not dictate the architecture of Python instance.

docs/software-installation-guide.md
ziw-liu: Installing MM is not a recorder issue. Relocating this discussion would be more appropriate.

talonchandler (approved):
I made minor rewording changes, added to the quickstart, and improved the tabs. Looks good from my perspective.

ziw-liu:
@talonchandler Sorry but I actually meant this ⬆️

talonchandler:
Roger that...looks great!

PR #167, opened 2022-09-08
Fix background correction for non-square images and online mode

talonchandler:
This PR fixes background correction bugs for:
- non-square cameras --- the `load_bg` function contained a bug that caused non-square images to apply a constant background correction instead of a pixel-wise background correction. This bug affected automaton and Yuming's setup (Oryx cameras), but not hummingbird which is why this bug slipped by previous tests. 
- online mode --- online results now match offline results in 'Estimated' and 'Measured + Estimated' modes. Currently, the online and offline modes use different section of the code for background correction, and this PR brings these sections into sync. 

I tested these fixes on automaton, and I will merge after code review. 

I'll test more thoroughly on hummingbird in advance of our 0.2.0 release.

### DIFF
--- a/recOrder/io/utils.py
+++ b/recOrder/io/utils.py
@@ -72,7 +72,7 @@ def load_bg(bg_path, height, width, ROI=None):
     for i in range(len(bg_paths)):
         img = tiff.imread(bg_paths[i])
 
-        if ROI is not None and ROI != (0, 0, height, width):
+        if ROI is not None and ROI != (0, 0, width, height):
             bg_data[i, :, :] = np.mean(img[ROI[1]:ROI[1] + ROI[3], ROI[0]:ROI[0] + ROI[2]])
 
         else:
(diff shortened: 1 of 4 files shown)


recOrder/io/utils.py
```diff
@@ -72,7 +72,7 @@ def load_bg(bg_path, height, width, ROI=None):
     for i in range(len(bg_paths)):
         img = tiff.imread(bg_paths[i])
 
-        if ROI is not None and ROI != (0, 0, height, width):
```
talonchandler, line 75: This was the key offending line that caused non-square images to fail.

recOrder/pipelines/qlipp_pipeline.py
```diff
@@ -72,13 +72,9 @@ def __init__(self, config: ConfigReader, data: WaveorderReader, writer: Waveorde
...
 
-        # Convert 'local_fit+' to 'local_fit' for waveorder
-        if self.config.background_correction == 'local_fit+':
-            wo_background_correction = 'local_fit'
-        else:
-            wo_background_correction = self.config.background_correction
+        wo_background_correction = rec_bkg_to_wo_bkg(self.config.background_correction)
 
-        # Initialize Reconstructor
+         # Initialize Reconstructor
```
ziw-liu, line 77: Extra space here 🔎

ziw-liu (approved):
Code-wise it seems reasonable. More testing with hardware is needed to confirm.

PR #169, opened 2022-09-11
Remove background averaging 

talonchandler:
When I was testing calibration using an ROI, I noticed that the background correction changed behavior. 

I found that when a calibration ROI was used, the background correction averaged over this ROI then applied that background correction uniformly over the whole image. I cannot think of a reason to average over a FOV within a background image, so I removed this functionality. 

This functionality severely reduces the effectiveness of the "Measured" background corrections when calibrating within an ROI. It's also likely that this bug was partially responsible for Yuming's poor background corrections.

### DIFF
--- a/recOrder/io/utils.py
+++ b/recOrder/io/utils.py
@@ -52,14 +52,13 @@ def extract_reconstruction_parameters(reconstructor, magnification=None):
     return attr_dict
 
 
-def load_bg(bg_path, height, width, ROI=None):
+def load_bg(bg_path, height, width):
     """
     Parameters
     ----------
     bg_path         : (str) path to the folder containing background images
     height          : (int) height of image in pixels
     width           : (int) width of image in pixels
-    ROI             : (tuple)  ROI of the background images to use, if None, full FOV will be used
 
     Returns
     -------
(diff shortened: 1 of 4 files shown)


ziw-liu:
Do you think a warning is still needed since we are altering the behavior of the pipeline?

talonchandler:
I *think* I removed the warning and decided to completely remove the background averaging. 

I initially thought I was going to just add a warning, then I changed my mind and decided to remove background averaging. Sorry I didn't do a very good job cleaning up after myself: the warning is still in the commit history and in the name of this branch.

ziw-liu:
> Sorry I didn't do a very good job cleaning up after myself: the warning is still in the commit history and in the name of this branch.

I wasn't making it clear, but I was suggesting adding the warning back (with different articulation) so that any user who was relying on that bug/feature would be advised of the change we are making. Although the previous behavior was more like a bug, this 'breaking' change still alters the output value of the pipeline if, for example, someone would like to revisit their previous analysis.

talonchandler:
Aha! Excellent point. I didn't maintain backwards compatibility. 

I've reverted the changes, added a warning, fixed the bug, and added reminders for what needs to be removed for 1.0.0. 

Thanks Ziwen...great catch.

recOrder/io/utils.py
```diff
@@ -72,9 +72,12 @@ def load_bg(bg_path, height, width, ROI=None):
...
 
-        if ROI is not None and ROI != (0, 0, width, height):
-            bg_data[i, :, :] = np.mean(img[ROI[1]:ROI[1] + ROI[3], ROI[0]:ROI[0] + ROI[2]])
-
+        if ROI is not None and ROI != (0, 0, width, height): # TODO: Remove for 1.0.0
+            warning_msg = """
+            Warning: earlier versions of recOrder would have averaged over the background ROI. This behavior is 
+            now considered a bug, and future versions of recOrder will not average over the background. 
+            """
+            print(warning_msg)
```
ziw-liu, line 80: Same question as in [#157](https://github.com/mehta-lab/recOrder/pull/157#discussion_r968971351), as this message may not be visible from the napari window.

ziw-liu (approved):

recOrder/io/utils.py
talonchandler: Again...good catch. Changed from `print` to `logging.warning`.

PR #171, opened 2022-09-14
Clean dependencies

talonchandler:
@ziw-liu Following our discussion, I've removed a couple dependencies:

- grip (we can use an IDE instead)
- pyqt5
- PyQT5 -> qtpy (to support PySide2 following napari [best practices](https://napari.org/stable/plugins/best_practices.html). 

Changing to qtpy required a global find+replace of `pyqtSignal` -> `Signal`, and `pyqtSlot` -> `Slot`.

Feel free change other dependencies on this branch. Thanks!


recOrder/plugin/qtdesigner/recOrder_ui.py
```diff
@@ -2,13 +2,13 @@
 
 # Form implementation generated from reading ui file '.\recOrder_ui.ui'
 #
-# Created by: PyQt5 UI code generator 5.15.4
+# Created by: qtpy UI code generator 5.15.4
```
ziw-liu, line 5: Not sure if we should alter this since it seems to be auto-generated.

ziw-liu (approved):

PR #173, opened 2022-09-15
Set focus to "Plot" tab by default

talonchandler:
Almost every time I run a calibration I need to scroll down and click the "Plot" tab. This changes the default focus.


ziw-liu (approved):
Works fine locally and on Automaton!

PR #174, opened 2022-09-15
Show RAM warning in GUI / log in CLI

ziw-liu:
The RAM implementation in [#157](https://github.com/mehta-lab/recOrder/pull/157#discussion_r972256905) does not offer a visual hint in the GUI. The use of `print()` in the CLI also limited the way we can format the message. Here is a modified version with a more portable API.

### DIFF
--- a/recOrder/io/utils.py
+++ b/recOrder/io/utils.py
@@ -172,18 +172,33 @@ def get_unimodal_threshold(input_image):
     assert best_threshold > -np.inf, 'Error in unimodal thresholding'
     return best_threshold
 
+
 def ram_message():
+    """
+    Determine if the system's RAM capacity is sufficient for running reconstruction. 
+    The message should be treated as a warning if the RAM detected is less than 32 GB.
+
+    Returns
+    -------
+    ram_report    (is_warning, message)
+    """
     BYTES_PER_GB = 2**30
     gb_available = psutil.virtual_memory().total / BYTES_PER_GB
-    if gb_available < 32:
-        return ' \n'.join(textwrap.wrap(
-               f'\033[91mWARNING:\033[0m recOrder reconstructions often require more than the {gb_available:.1f} ' \
+    is_warning = gb_available < 32
+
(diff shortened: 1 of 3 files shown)


talonchandler:
Look good...I'm on board. 

Would you mind testing this before merging? I think it will be easiest to test locally or on the HPC where you can control how much RAM is available.

ziw-liu:
> Would you mind testing this before merging? I think it will be easiest to test locally or on the HPC where you can control how much RAM is available.
I tested locally (16 GB RAM) and on the Automaton computer (128 GB RAM) and it works as expected.
Not sure about the CLI part yet.

ziw-liu:
@talonchandler can you help test on the CLI to see if the message is displayed properly?

talonchandler:
@ziw-liu When you tested the napari `showwarning` function, did your warning messages persist in napari? 

I'm finding that as I step through `showwarning` I get the correct warning message, but when I run it normally the message doesn't persist.

ziw-liu:
> did your warning messages persist in napari?

I think napari manages the persistence of the message. In my testing it'll only stay on for a few seconds but will persist if you click the expand icon.

talonchandler:
I just tested the CLI and the message appears as expected. 

I also changed the behavior so that the message appears in napari when you run the pipeline in offline mode.

talonchandler (approved):

ziw-liu:
> I also changed the behavior so that the message appears in napari when you run the pipeline in offline mode.
Will the method ever be called when there is not a napari instance (e.g. in CLI mode)? That will cause an error.

talonchandler:
Have you tested? I thought it would error, too, but I tried it and I didn't get an error.

talonchandler:
If you haven't run the CLI or offline mode yet, I can give you a quick intro/tutorial. It can be a bit finicky, but it works pretty well.

ziw-liu:
> Have you tested?
No I haven't, just speculating 🤔

PR #175, opened 2022-09-15
Improved warning message

talonchandler:
- doesn't print 4/5 times
- doesn't repeat "WARNING" (thanks ziwen)

### DIFF
--- a/recOrder/io/utils.py
+++ b/recOrder/io/utils.py
@@ -72,17 +72,16 @@ def load_bg(bg_path, height, width, ROI=None):
     bg_paths.sort()
     bg_data = np.zeros([len(bg_paths), height, width])
 
+    if ROI is not None and ROI != (0, 0, width, height): # TODO: Remove for 1.0.0
+        warning_msg = """
+        Earlier versions of recOrder would have averaged over the background ROI. This behavior is 
+        now considered a bug, and future versions of recOrder will not average over the background. 
+        """
+        logging.warning(warning_msg)
+
     for i in range(len(bg_paths)):
         img = tiff.imread(bg_paths[i])
-
-        if ROI is not None and ROI != (0, 0, width, height): # TODO: Remove for 1.0.0
-            warning_msg = """
-            Warning: earlier versions of recOrder would have averaged over the background ROI. This behavior is 
-            now considered a bug, and future versions of recOrder will not average over the background. 
-            """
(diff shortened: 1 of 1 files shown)


recOrder/io/utils.py
```diff
@@ -72,17 +72,16 @@ def load_bg(bg_path, height, width, ROI=None):
...
-
-        if ROI is not None and ROI != (0, 0, width, height): # TODO: Remove for 1.0.0
-            warning_msg = """
-            Warning: earlier versions of recOrder would have averaged over the background ROI. This behavior is 
-            now considered a bug, and future versions of recOrder will not average over the background. 
-            """
-            logging.warning(warning_msg)
-        else:
-            bg_data[i, :, :] = img
+        bg_data[i, :, :] = img
```
ziw-liu, line 84: This line previously had more indentation (under an `else` clause, only executed if using an ROI).

talonchandler: Oh you're 100% right! Nice catch...this would have caused bugs.

ziw-liu: `bg_data` value assignment when `ROI` is used was removed in [this commit](https://github.com/mehta-lab/recOrder/pull/169/commits/dbbe800aa5f3e346e8bf3cd52b419e70232133ed).

talonchandler: On a closer look, I think I'm happy with the current implementation. 

The last stable release would always return a background image...sometimes an averaged one. I think the `load_bg` function should only load the background...no averaging or other operations. 

So I have changed the behavior of `load_bg` to always load the background. If the previous implementation would have averaged, I print a warning message to let the user know that I've broke backwards compatibility (even though I would call the previous behavior a bug).

talonchandler: By "current implementation" I mean the proposed PR in this branch.

ziw-liu: If the behavior of the current PR will differ from main (which then differs from 0.1.2), should we clarify in the warning message with specific version numbers? e.g. `Earlier versions of recOrder` -> `recorder 0.1.2 and lower`, and `future versions` -> `current and future versions`.

talonchandler: Great call. I just added a commit that specifies which versions I'm talking about.

ziw-liu (approved):

PR #177, opened 2022-09-16
Hide "listen" and "overlay" features from `recOrder` GUI

talonchandler:
See screenshots of the changes below:
Screenshot 2022-09-16 113632

image

This PR is consistent with our goal for 0.2.0 --- all user-facing features should work well. I'm hiding these features because they are not currently working well, and we're planning to fix/improve them for 1.0.0.

### DIFF
--- a/recOrder/plugin/widget/main_widget.py
+++ b/recOrder/plugin/widget/main_widget.py
@@ -301,6 +301,15 @@ class MainWidget(QWidget):
         self.ui.label_focus_zidx.setHidden(True)
         self.ui.le_focus_zidx.setHidden(True)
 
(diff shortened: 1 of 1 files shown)


talonchandler:
@Soorya19Pradeep have you been using the "overlay" features at all? I want to confirm that this change doesn't affect you negatively.

Soorya19Pradeep:
> @Soorya19Pradeep have you been using the "overlay" features at all? I want to confirm that this change doesn't affect you negatively.

@talonchandler , I don't use the overlay feature.

ziw-liu (approved):

PR #178, opened 2022-09-16
`Capture Background` without calibration yields a warning

talonchandler:
Currently, clicking `Capture Background` before loading or running a calibration gives the following error:
`Error: 'NoneType' object has no attribute 'capture_bg'`

This PR changes `recOrder`'s behavior so that the following warning is shown instead: 

```Capturing a background requires calibrated liquid crystals. Please either run a calibration or load a calibration from file.```

(Please excuse the extra commits---I'm still getting the hang of VScode :) )

### DIFF
--- a/recOrder/plugin/widget/main_widget.py
+++ b/recOrder/plugin/widget/main_widget.py
@@ -1,5 +1,3 @@
-from multiprocessing.sharedctypes import Value
-from tkinter.messagebox import showerror
 from recOrder.calib.Calibration import QLIPP_Calibration, LC_DEVICE_NAME
 from pycromanager import Bridge
 from qtpy.QtCore import Slot, Signal, Qt
--- a/recOrder/plugin/widget/main_widget.py
+++ b/recOrder/plugin/widget/main_widget.py
@@ -301,6 +301,15 @@ class MainWidget(QWidget):
         self.ui.label_focus_zidx.setHidden(True)
         self.ui.le_focus_zidx.setHidden(True)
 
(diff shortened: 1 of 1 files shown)


ziw-liu:
I'm not sure if flashing a warning and returning `None` early is the best way to handle this. We can also consider checking the input at `BackgroundCaptureWorker.__init__()` and raising an `ValueError` if `None` is given as `calib`. This is more expensive (to spawn a worker thread before aborting) but I feel like leaving it to the worker makes it more decoupled.

talonchandler:
I'm fine with either. I chose to spawn the warning in the main thread to make the warning/failure happen as fast as possible, but if you think there's a code coupling tradeoff that I'm not thinking about then I'm happy to move it to the worker. 

FWIW there are many other validation checks happening in `main_widget.py` (for better or worse). For example, `_check_requirements_for_acq` checks inputs before shipping to a thread. 

I'm not too concerned about our short-term choice here, but I definitely think this is worth thinking about for our long-term solution.

ziw-liu:
@talonchandler
What's your thought on error vs warning?

talonchandler:
Hmmm not sure. I just googled around and reminded myself of a rule I've heard before: "if you can recover use a warning, if not use an error". I guess `recOrder` can't really recover here w/o the user's help, so error?

I'm open to different error vs. warning rules, too.

talonchandler:
I changed the warning to an error and tested on hummingbird. This is ready to merge from my perspective.

ziw-liu (approved):

PR #183, opened 2022-09-20
Upgrade `pycromanager` and `micromanager` 

talonchandler:
This PR upgrades:

`pycromanager == 0.13.2 --> 0.19.2`
`micromanager == 20210713 --> 20220901`.

### DIFF
--- a/docs/microscope-installation-guide.md
+++ b/docs/microscope-installation-guide.md
@@ -27,11 +27,11 @@ should launch napari (may take 15 seconds on a fresh installation) with the recO
  
 ## Install and configure `Micromanager`
 
-Install `Micromanager 2.0` nightly build `20210713` (https://micro-manager.org/Micro-Manager_Nightly_Builds). 
+Install `Micromanager 2.0` nightly build `20220901` (https://micro-manager.org/Micro-Manager_Nightly_Builds). 
 
-**Note:** We have tested recOrder with `20210713`, but most features will work with newer builds. We recommend testing a minimal installation with `20210713` before testing with a different nightly build or additional device drivers. 
+**Note:** We have tested recOrder with `20220901`, but most features will work with newer builds. We recommend testing a minimal installation with `20220901` before testing with a different nightly build or additional device drivers. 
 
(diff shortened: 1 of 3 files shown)


talonchandler:
This upgrade went smoothly. I've tested on hummingbird, and it's ready to merge from my perspective. 

@ieivanov @ziw-liu after this merges you will need to use micromanager 20220901 and a fresh install of `recOrder` so that `pycromanager` gets upgraded (or you can try `pip install -U pycromanager` within your existing environment).

talonchandler:
I removed one line from the docs---no need to link to a specific release anymore.

ziw-liu (approved):

docs/microscope-installation-guide.md
```diff
@@ -27,11 +27,11 @@ should launch napari (may take 15 seconds on a fresh installation) with the recO
...
+**Note:** We have tested recOrder with `20220901`, but most features will work with newer builds. We recommend testing a minimal installation with `20220901` before testing with a different nightly build or additional device drivers. 
 
-Before launching `Micromanager`, download the Meadowlark device adapters and calibration files from the [release page](https://github.com/mehta-lab/recOrder/releases/) or [directly for 0.1.0](https://github.com/mehta-lab/recOrder/releases/download/0.1.0/recOrder-0.1.0-device-drivers.zip) and place these three unzipped files into your `Micromanager` folder (likely `C:\Program Files\Micro-Manager-2.0.1_2021_07_13` or similar). 
+Before launching `Micromanager`, download the Meadowlark device adapters and calibration files from the [release page](https://github.com/mehta-lab/recOrder/releases/) and place these three unzipped files into your `Micromanager` folder (likely `C:\Program Files\Micro-Manager-2.0.1_2022_09_` or similar).
```
ieivanov, line 34: I think we can replace the MM folder name with "`C:\Program Files\Micro-Manager` or similar" to keep the instructions simple and a bit easier to maintain

ieivanov:
With this update of pycromanager, I believe calls like `mmc = bridge.get_core()` raise a deprecation warning. We should change this syntax to 

```python
from pycromanager import Core
mmc = Core()
```

docs/microscope-installation-guide.md
talonchandler: Good call. Fixed.

talonchandler:
Thanks for bringing this up, Ivan. 

I just tried changing to the recommended `from pycromanager import Core; mmc = Core()`, and I ran into a snag: `Core()` doesn't seem to support `convert_camel_case=False`, so if we make this change it seems we'll need to change all of our `Core` and `Studio` calls from CamelCase to snake_case. 

I vote that we leave it as is for now, since `bridge.get_core()` seems softly deprecated---`pycromanager`'s tests and examples still use `bridge.get_core()`. If it becomes truly deprecated/unsupported, we can make the change for 1.0.0 when our dependence on `pycromanager` will be much weaker and we'll have far fewer places to change and manually test.

Issue #180, opened 2022-09-16
[BUG] "Load calibration" followed by "Capture Background" gives error


talonchandler:
"Load calibration" followed by "Capture Background" yields:
`Error: unsupported operand type(s) for *: 'NoneType' and 'float'`

talonchandler:
It seems like the `lca{a/b}_{0,60,120}` attributes of `QLIPP_Calibration` are not being set correctly during "Load calibration".

Ziwen, would you mind helping out with this? I think you should be able to recreate this issue locally.

ziw-liu:
Found these lines in `load_calibration`:

https://github.com/mehta-lab/recOrder/blob/00cf5ad0d733aad44067a8e65639f67abb446cd5/recOrder/plugin/workers/calibration_workers.py#L370-L372

It seems to me that the `return` line does not do anything here?

talonchandler:
I *think* the `return` is used, but in a sneaky way by `self.worker.returned` here: 

https://github.com/mehta-lab/recOrder/blob/main/recOrder/plugin/widget/main_widget.py#L2460

It seems like the calibration must be returned somehow, or else the LCs wouldn't receive any of the data from the thread. 

This page was pretty helpful for me to understand how this section works:
https://napari.org/stable/guides/threading.html

"The [@thread_worker](https://napari.org/stable/api/napari.qt.threading.html#napari.qt.threading.thread_worker) decorator converts your function into one that returns a [WorkerBase](https://napari.org/stable/api/napari.qt.threading.WorkerBase.html#napari.qt.threading.WorkerBase) instance."

PR #184, opened 2022-09-20
Initialize calibration values when loading from metadata files

ziw-liu:
Previously loading calibration from a metadata file does not initialize LC values in the `QLIPP_Calibration` object created by the `load_calibration` thread worker, which [causes an error when trying to capture background afterwards](https://github.com/mehta-lab/recOrder/issues/180). This PR addresses the error by setting the required values, although the necessity of fully initializing the `QLIPP_Calibration` object when loading from metadata files is still up to debate.

### DIFF
--- a/recOrder/plugin/workers/calibration_workers.py
+++ b/recOrder/plugin/workers/calibration_workers.py
@@ -373,7 +373,7 @@ def load_calibration(calib, metadata: MetadataReader):
                 # set the swing value attribute (e.g. 'swing0')
                 if state != "ext":
                     swing_name = "swing" + state
-                    setattr(calib, swing_name, metadata.Swing_measured[i + 1])
+                    setattr(calib, swing_name, metadata.Swing_measured[i - 1])
 
     _set_calib_attrs(calib, metadata)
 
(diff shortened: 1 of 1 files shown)


ziw-liu:
I'm not currently labelling this issue as automatically closing #180 upon merging. Feel free to change it as needed.

recOrder/plugin/workers/calibration_workers.py
```diff
@@ -1,3 +1,4 @@
+from argparse import MetavarTypeHelpFormatter
```
talonchandler, line 1: Does this do anything?

talonchandler (approved):
Thanks @ziw-liu! 

I just tested, and everything is working as I expect. I will link #180 since this fixes that issue. 

My only question is about the new import statement. Is it necessary?

As we discussed, the data flow in the calibration metadata files is a bit wonky right now. We'll design this more carefully and clean this up for 1.0.0. 

Thanks!

recOrder/plugin/workers/calibration_workers.py
ziw-liu: Oops that's an unintended behavior of my editor 😅

PR #187, opened 2022-09-21
installation of recorder on osx_arm64

mattersoflight:
See closed PR mehta-lab/recOrder#186.


talonchandler:
Thanks Shalin!

I just naively tested
`conda env create -f conda_env_OSX_arm64.yml`
on my M1 and received `Solving environment: failed` and `ResolvePackageNotFound:...`, 
so it seems like there need to be additional instructions beyond a simple "use this .yml-based environment".

FWIW when I type `arch` from my terminal I get `arm64`. 

I'll work with Ziwen to get to get this solved in a way that works out of the box from our M1s, and we'll likely ask you to test before merging.

ziw-liu:
I just verified with @edyoshikun and his fresh M1 Mac. It seems that the 'official' `conda` from Anaconda would be `x86_64` by default, downloading packages from `osx64`. And if one run `time.sleep(time_to_sleep)` in Python the 'Python 3.x' process would be 'Intel' in the Acitivity Monitor. In the meantime, the native `conda` (e.g. from conda-forge) would default to `arm64` (downloading package from `osx-arm64`). This explains the reason why @mattersoflight, @Christianfoley and I see the memory leak when doing `pip install pyqt5` (required by napari), but @talonchandler has never encountered such an issue.

talonchandler:
Great sleuthing, Ziwen!

> forcing a sub-channel on native conda installs should be sufficient to resolve https://github.com/mehta-lab/recOrder/issues/166.

Does this mean that specifying a channel and/or subchannel when we create the environment would be enough? Something like `conda create -n <env-name> -c <channel-name> python=3.9`? That would be excellent.

ziw-liu:
> Something like `conda create -n <env-name> -c <channel-name> python=3.9`?

Both `CONDA_SUBDIR=osx-64 conda create -n <env-name> python=3.9` and `conda create -n <env-name> -c conda-forge/osx-64 python=3.9` seems to work on my end. The later would require the user adding the `conda-forge` channel if they do not already have that. Can you test the first one to see it it works with the 'official' Anaconda `conda`?

ziw-liu:
> Can you test the first one to see it it works with the 'official' Anaconda `conda`?

I just tested with @edyoshikun and this works. Now this seems like the most universal solution to me.

ziw-liu:
The current commit of this PR is no longer a working solution after we explicitly depend on `napari[all]` in `setup.cfg`. `pip` will still attempt to install `napari` and thus `PyQt5` even `conda` has already installed a native version, and cause a 'quiet' memory leak.

ziw-liu:
Partially resolved with #165

ziw-liu:
@mattersoflight I think we can close this PR given https://github.com/mehta-lab/recOrder/issues/193#issuecomment-1319382740

ziw-liu:
Closing since the branch is stale and the issue is mostly resolved without the need of using a conda env file. @mattersoflight Can I also delete the branch from GitHub to keep the number of active remote branches down?

PR #196, opened 2022-09-26
Mapping capture background path from calibration to reconstruction

edyoshikun:
Added Signal/Slot to update the acquired background path on the reconstruction from the background capture button in acquisition.

### DIFF
--- a/recOrder/plugin/widget/main_widget.py
+++ b/recOrder/plugin/widget/main_widget.py
@@ -2113,7 +2113,7 @@ class MainWidget(QWidget):
             self.ui.le_bg_path.setText('Path Does Not Exist')
 
     @Slot(str)
-    def handle_bg_path_update(self, value:str):
+    def handle_bg_path_update(self, value: str):
         """
         Handles the update of the most recent background folderpath from
         BackgroundWorker to display in the reconstruction texbox.
--- a/recOrder/plugin/widget/main_widget.py
+++ b/recOrder/plugin/widget/main_widget.py
@@ -2112,6 +2112,17 @@ class MainWidget(QWidget):
         else:
             self.ui.le_bg_path.setText('Path Does Not Exist')
 
+    @Slot(str)
+    def handle_bg_path_update(self,value):
+        #Path from Background Worker
+        path = value
+        if os.path.exists(path):
+            self.acq_bg_directory = path
+            self.current_bg_path = path
+            self.ui.le_bg_path.setText(path)
+        else:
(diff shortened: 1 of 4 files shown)


edyoshikun:
This was using Falcon, for which we updated mmc version 09012022 and the respective components to work with RecOrder.

recOrder/plugin/widget/main_widget.py
```diff
@@ -2112,6 +2112,17 @@ def enter_acq_bg_path(self):
         else:
             self.ui.le_bg_path.setText('Path Does Not Exist')
 
+    @Slot(str)
+    def handle_bg_path_update(self,value):
```
ziw-liu, line 2116: Consider PEP 8 issue: `(self,value)` -> `(self, value)`.

ziw-liu: Also please include docstrings for your public methods.

recOrder/plugin/widget/main_widget.py
```diff
@@ -2112,6 +2112,17 @@ def enter_acq_bg_path(self):
...
+        #Path from Background Worker
+        path = value
+        if os.path.exists(path):
+            self.acq_bg_directory = path
+            self.current_bg_path = path
+            self.ui.le_bg_path.setText(path)
+        else:
+            self.ui.le_bg_path.setText('Path Does Not Exist')
```
ziw-liu, line 2124: Refer to: https://github.com/mehta-lab/recOrder/blob/f4d6407ca7d226ffddf04bfc77afc31d677c3aa5/recOrder/plugin/widget/main_widget.py#L2276-L2277

ziw-liu: Ideally we also want a napari-level warning banner for all these events. Maybe worth a separate issue for 1.0.0+.

ziw-liu: On second thought, do we need this if the parameter is only passed in from within `recOrder`? How about a try/except instead?

recOrder/plugin/widget/main_widget.py
edyoshikun: Thank you! I'll be mindful for future PRs.

ziw-liu (approved):
Looks good to me -- great first PR!

talonchandler:
Just tested on hummingbird---works extremely well. 

Great first PR, Ed!

talonchandler (approved):

PR #199, opened 2022-09-28
Plot calibrated LC states

### LINKED ISSUE #323: [FEATURE] Add LCA vs. LCB plot of calibrated states
Plotting the calibrated LCA/LCB states can help us identify calibration issues---see mehta-lab/recOrder#194 for an example. This feature request is for an additional LCA/LCB plot in `recOrder` that shows the user where the calibrated states are in LCA/LCB space. 

The plot needs to:
- appear somewhere in the calibration tab...maybe beside the intensity plot or in a separate subtab if the intensity plot is already too small. 
- plot all 4 or 5 LCA/LCB points 
- have labelled axes "LCA retardance" and "LCB retardance" and points ("ext", "0", "45", "90", "135", etc)
- zooming and panning like the current intensity plot

It would be nice to have:
- each LCA/LCB point populate as soon as it's done (maybe a user could spot an early failure and abort?)
- good static default xlims and ylims. You should be able to center the plot around the extinction state with radius of ~1.5*swing. 
- a saved LCA/LCB plot alongside the calibration metadata? We've also sometimes wanted to revisit the intensity trace, so it might be nice to have this saved too. 

@ziw-liu this would be a huge improvement over the way we've been debugging currently. You can decide about the "nice to haves" depending on how much additional work they'll take. Thank you!

ziw-liu:
As mehta-lab/waveorder#323 suggests, plotting the calibrated LC states can be useful for debugging mehta-lab/waveorder#324 and validating calibration in general. This PR approaches the feature request in a minimal and contained way without tampering with any existing UI component.

**Expected behavior**
A pop-up window appears upon the finishing of a calibration job:
win_10_plot_demo

**Concerns**
A pop-up window is not the most elegant design and is against the `napari` philosophy.

### DIFF
--- a/recOrder/calib/Calibration.py
+++ b/recOrder/calib/Calibration.py
@@ -921,7 +921,7 @@ class QLIPP_Calibration():
             "axes.spines.top": False,
         }):
             fig = plt.figure("Calibrated LC States")
-            plt.scatter(lc_values["LCA"], lc_values["LCB"], 'r')
+            plt.scatter(lc_values["LCA"], lc_values["LCB"], c='r')
             plt.axis("equal")
             plt.xlabel("LCA retardance")
             plt.ylabel("LCB retardance")
--- a/recOrder/calib/Calibration.py
+++ b/recOrder/calib/Calibration.py
@@ -919,7 +919,7 @@ class QLIPP_Calibration():
         with plt.rc_context({
             "axes.spines.right": False,
             "axes.spines.top": False,
-            "axis.axis": "equal",
+            "axes.axis": "equal",
         }):
             fig = plt.figure("Calibrated LC States")
             plt.scatter(lc_values["LCA"], lc_values["LCB"], 'ro')
--- a/recOrder/calib/Calibration.py
+++ b/recOrder/calib/Calibration.py
@@ -911,7 +911,7 @@ class QLIPP_Calibration():
         lc_sides = ("A", "B")
         lc_values = {
             "LC{lc_side}": [
-                self.__getattribute__("lc" + lc_side + pol) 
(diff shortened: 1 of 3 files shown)


ziw-liu:
Just realized that `matplotlib` isn't thread safe. Reworking the implementation...

ziw-liu:
Now working as (I) expected.

talonchandler:
Can I request that the axis appears square with the same x and y widths? Apologies for the nit pick, but I think it will help spot issues.

ziw-liu:
> the axis appears square with the same x and y widths

Can you clarify on what aspect ratio you need? I think I've already copied the `axis("equal")` call from your plotting script.

talonchandler:
Got it...my bad. I mistakenly though the rectangular plotting area was a problem. 

I'll test this later today. Thanks again.

talonchandler:
I just added a quick circle plot to make it easier to gauge non-circular results. See below: 

image

It's weird that the spines weren't hidden properly on hummingbird. Possibly windows? Not worth worrying about IMO. 

LGTM! I'll let you merge after you've looked at my changes. Thanks Ziwen.

talonchandler (approved):

talonchandler:
Actually...hang on. I may have found an issue with repeated calibrations. Let me test for a few more minutes.

talonchandler:
I found that if you didn't close the previous plot then a new calibration plot wouldn't appear. To fix it I closed all open plots before starting the new plot.

I also fixed the fixed annotation width...for small swings the labels drifted away from the point. 

Now I think this is ready. Go ahead and test + merge. Thanks!

talonchandler:
For reference, here's what the plot looks like when the calibration fails:
image

The circle makes it much clearer what's going on imo.

PR #202, opened 2022-09-28
Remove pre-acquisition background file check for 'Estimated' background correction

### LINKED ISSUE #200: [BUG] Acquiring with an "Estimated" background correction fails if "Background Path" is empty
Yuming pointed out a bug: if you press "Acquire Birefringence" with "Estimated" background correction and the "Background Path" field is empty, it will fail. The tab become red and the hidden "Background Path" field also becomes red. You can fix this by changing the "BG Correction" box to "Measured", entering a path into the field, then changing back to "Estimated", but this is a kludge. 

@ziw-liu can you fix this so that opening recOrder and clicking "Acquire Birefringence" with "Estimated" works right away? Thanks for your help!

ziw-liu:
Fixes #200.

### DIFF
--- a/recOrder/plugin/widget/main_widget.py
+++ b/recOrder/plugin/widget/main_widget.py
@@ -779,7 +779,8 @@ class MainWidget(QWidget):
                 raise_error = True
                 self._set_tab_red('General', True)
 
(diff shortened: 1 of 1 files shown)


talonchandler (approved):
Tested and everything works well! Thanks.

PR #208, opened 2022-09-29
Add "pad_z" when constructing microscope metadata

### LINKED ISSUE #207: "Microscope Parameters" mismatch in writing and loading calibration metadata
**Bug**
`"pad_z"` is an expected field in [`MainWidget.load_calibration`](https://github.com/mehta-lab/recOrder/blob/d032484c255615796ba9121ad3b049bb289e8817/recOrder/plugin/widget/main_widget.py#L2478), but not specified in [`CalibrationWorkerBase._microscope_params`](https://github.com/mehta-lab/recOrder/blob/d032484c255615796ba9121ad3b049bb289e8817/recOrder/plugin/workers/calibration_workers.py#L101-L107). This causes a `KeyError` when loading a previous calibration.

**Reproduction**
Click 'LC Calibration -> Run Calibration -> Load Calibration)' in the `recOrder` GUI and select a calibration metadata file.

ziw-liu:
Fixes #207

### DIFF
--- a/recOrder/plugin/widget/main_widget.py
+++ b/recOrder/plugin/widget/main_widget.py
@@ -2475,7 +2475,6 @@ class MainWidget(QWidget):
         # Update the Microscope Parameters with those from the previous calibration (if they're present)
         params = metadata.Microscope_parameters
         if params is not None:
-            self.ui.le_pad_z.setText(str(params['pad_z']) if params['pad_z'] is not None else '')
             self.ui.le_n_media.setText(str(params['n_objective_media']) if params['n_objective_media'] is not None else '')
             self.ui.le_obj_na.setText(str(params['objective_NA']) if params['objective_NA'] is not None else '')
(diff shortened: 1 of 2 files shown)


talonchandler:
Thanks for spotting this bug, Ziwen. 

I accidentally introduced this in #163 when I removed `pad_z`, `bg_correction_option` from the `Microscope params` since these settings aren't parameters of the microscope, they're parameters for reconstruction. 

I also removed `swing` from `Microscope params`, but left it among the high-level calibration metadata fields---it was being written twice.  

So I would vote that the fix for #207 should be to remove `pad_z` from the expected parameters instead of returning it to the written parameters. `pad_z` should (and is AFAIK) be written and read from the config files, not the calibration metadata files. 

Thanks!

ziw-liu:
> the fix for https://github.com/mehta-lab/recOrder/issues/207 should be to remove pad_z from the expected parameters

Done!

talonchandler (approved):

PR #212, opened 2022-09-30
Fix bug in calibration_metadata writer

ieivanov:
Changing `V_min` and `V_max` to `float` as  `QLIPP_Calibration.write_metadata` fails to write json metadata if integer voltage is passed.


ziw-liu (approved):
Looks good!

PR #214, opened 2022-10-01
Fix silent calibration failures

talonchandler:
The following figure summarizes the fix in this PR:

pr-plot

**Caption:** before this PR the calibrations were being optimized over a region of fixed width = 0.05 waves. For small swing values (swing = 0.04 pictured here), the optimization region crossed the extinction state which often resulted in calibrated states on the wrong side of the extinction state. 

This PR scales the optimization with the swing so that the calibrated states are always on the correct side of the extinction state. 

**Details:** I've changed the LCA search region from a fixed `0.05` to `min(swing/ratio, 0.05)` for 4-state and `min(swing, 0.05)` for 5-state calibrations. This means that the behavior for large swings (>0.05 for 5-state and >0.1 for 4-state) will be unchanged---only small-swing calibrations are affected by this PR.

**More data:** this PR also includes an automation script that I used to collect 40 calibrations in all combinations of:
- [4, 5] state
- [0.1, 0.03, 0.01, 0.005] swing
- 5 repeats

Below are the LCA/LCB plots which show:
- that none of the calibrations failed---all of the states are placed on the correct side of the extinction state
- that the LC states vary by ~0.01 waves in LCA/LCB space. More work is needed to characterize the major contributors to this uncertainty. 

**Caption:** **Columns** correspond to decreasing swing left to right. **First 5 rows** correspond to repeats of the 4-state algorithm and the **last 5 rows** correspond to repeats of the 5-state algorithm. 
4-state
5-state


recOrder/scripts/calibrate.py
```diff
@@ -0,0 +1,30 @@
+# This script can be modified to debug and test calibrations
```
ziw-liu, line 1: Is this script best suited in the `recOrder` package (thus released with source code on PyPI)?

talonchandler: Fair point. A possible counterargument is that we release unit tests (although they run automatically whereas these have much stronger dependencies). 

What do you think about moving this script (and others I generated today) to `compmicro-sandbox`?

mattersoflight: Scripts to test calibration and acquisition can come be helpful when our collaborators run into bugs involving optical components and acquisition. Let's keep these scripts in `recOrder` repo. I'm fine with sharing the scripts with PyPI package. If we encounter a problem packaging these scripts, we can [exclude](https://stackoverflow.com/questions/45111877/exclude-a-folder-from-a-python-package-for-uploading-a-package-to-pypi) the scripts folder.

recOrder/scripts/calibrate.py
```diff
@@ -0,0 +1,30 @@
...
+
+import napari
+import time
+import matplotlib.pyplot as plt
+from recOrder.plugin.widget.main_widget import MainWidget
+from recOrder.calib.Calibration import QLIPP_Calibration
+from recOrder.plugin.workers.calibration_workers import CalibrationWorker
+
+save_dir = "Q:\\Talon\\2022_09_30_calibration_5state\\"
```
ziw-liu, line 10: If we do decide to release this script, maybe clean this?

recOrder/scripts/calibrate.py
talonchandler: Roger that...let's include this script.

recOrder/scripts/calibrate.py
talonchandler: You bet! Cleaned.

ziw-liu (approved):
@talonchandler How do you feel about a more informative filename such as `debug_calibration.py` for the new script? Other than that I think we are good to merge.

talonchandler:
Good call...fixed.

ieivanov:
Thanks for putting in this work @talonchandler! I liked your `repeat_calibration.py` script. Seems like the `qbutton_gui_mode` button need a better name :)

As you say, from the repeated calibrations you did, I think it's clear that our calibration algorithm needs a bit more work. (2,4), (6,4), and (9,3) are hardly examples of good calibration. At this point, I don't have a clear sense of where these discrepancies may come from. Do you know if the optimization algorithm reached approximately equal intensities in all cases? One thing I've been suspicious about is whether we allow enough time for the LCs to settle into a new state. For example, in `opt_I0` we snap an image right after changing the LC state. 

https://github.com/mehta-lab/recOrder/blob/f7946b914c22de1e7ff3deb1b6d76c07e0590fb3/recOrder/calib/Calibration.py#L425-L426

`define_lc_state` calls `waitForDevice` or `waitForConfig`, but I'm not sure if these really give the LCs enough time to settle. I think the LCs are well equilibrated by that point - the last steps of the calibration algorithm don't change the intensity much, but we could try giving it a bit more time.

Issue #141, opened 2022-08-05
Upgrade pycromanager to 0.18.3


talonchandler:
For `1.0.0` we are planning to move `io` from `waveorder` to `recOrder`.

In [this `waveorder` PR](https://github.com/mehta-lab/waveorder/pull/79
), @ieivanov ran into a lazy-loading IO issue in his pycromanager reader that is solved in pycromanager 0.18.3. 

I'm tagging this here so that we:
- upgrade pycromanager for 1.0.0
-  remember to test the pycromanager reader when we upgrade pycromanager

ziw-liu:
Is this resolved in [#183](https://github.com/mehta-lab/recOrder/pull/183)? If so we can remove this from the 1.0.0 milestone (as it will ship with 0.2.0) and close the issue.

talonchandler:
Thanks for flagging @ziw-liu. I will be testing the pycromanager reader with Rachel tomorrow morning, so I'll close this when I've confirmed that the pycromanager reader is working for her. 

I'll change this issue to 0.2.0 so that I don't forget.

ieivanov:
I just tested MM 090122 - it does not yet implement an important TriggerScopeMM bug fix, which was merged shortly after: https://github.com/micro-manager/mmCoreAndDevices/pull/238. It may be worth pushing the MM nightly build version by a couple of days later

talonchandler:
Roger that...thanks for testing on the Triggerscope, Ivan. I'll bump to 2022-09-20 and test it before I merge this.

PR #215, opened 2022-10-01
Recommend MM 2022-09-20

talonchandler:
In #141 @ieivanov requested an slightly newer version of MM so that a [recent change to a device driver was included](https://github.com/micro-manager/mmCoreAndDevices/pull/238).
I tested `recOrder` with 2022-09-20 and all acquisition functionality is working. `pycromanager` did not require an upgrade. 

This PR updates the docs and warning messages only.

### DIFF
--- a/docs/microscope-installation-guide.md
+++ b/docs/microscope-installation-guide.md
@@ -27,9 +27,9 @@ should launch napari (may take 15 seconds on a fresh installation) with the recO
  
 ## Install and configure `Micromanager`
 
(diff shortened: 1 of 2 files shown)


mattersoflight (approved):
Thanks @talonchandler. The two builds are within a month - the chance of a hiccup is communication between MM and pycro-manager is negligible. We run into incompatibilty when the MM device adapter API or the zeromq versions change.

ziw-liu (approved):

ieivanov:
Quick note - 2022-09-20 supports writing to NDTiff! That doesn't matter here, but it's a cool feature

PR #223, opened 2022-10-05
Reduce "Value" in default overlay

talonchandler:
This PR multiplies the rescales the "Value" in the default birefringence overlay by 0.5. 

Below is a sweep of different defaults---on the left is the previous default, and in the center is the proposed default. 

sweep


mattersoflight:
I traced the bug to the code that computes the overlay. The overlays are as expected after the fix.
image

recOrder/io/utils.py
```diff
@@ -283,30 +301,44 @@ def ret_ori_overlay(retardance, orientation, scale, mode='2D', cmap='JCh'):
...
     for i in range(slices):
-        ret_ = np.interp(retardance[i], ret_scale, scale)
-        ori_binned = np.round(orientation[i] / 180 * levels + 0.5) / levels - 1 / levels
-        ori_ = np.interp(ori_binned, (0, 1), (0, 360))
-
-        if cmap == 'JCh':
+        ret_ = np.copy(
+            retardance[i]
+        )  # copy to avoid modifying original array.
+        ret_[ret_ > ret_max = 10] = ret_max = 10  # clip retardance to specified max.
```
ziw-liu, line 329: Tests didn't pass due to a `SyntaxError` here.

ziw-liu: I (think I) fixed it!

talonchandler:
Thanks for the much cleaner fix @mattersoflight. 

For reference, 
I just tested on hummingbird and received the following overlay:
Screenshot 2022-10-05 132426

Merging now.

PR #226, opened 2022-10-05
Fix 'Load Default Config' in offline mode

ziw-liu:
The 'Load Default Config' button no longer initiate values from `ConfigReader`, but read a config file at `recOrder/plugin/config_offline_default.yml`.


ziw-liu:
Also please feel free to modify the 'default' values here. I generated these from the GUI startup values, so they are more like a set of 'example' values than solid 'default' settings.

PR #227, opened 2022-10-05
Remove microscope parameters from calibration metadata

### LINKED ISSUE #224: Remove "Microscope parameters" fields from `calibration_metadata.txt`
Splitting up #216. @ieivanov and I agreed that the "Microscope parameter" fields are not needed in `calibration_metadata.txt`.

ziw-liu:
Resolves #224.

### DIFF
--- a/recOrder/calib/Calibration.py
+++ b/recOrder/calib/Calibration.py
@@ -776,7 +776,7 @@ class QLIPP_Calibration():
 
             return inst_mat
 
-    def write_metadata(self, notes=None, microscope_params=None):
+    def write_metadata(self, notes=None):
 
         inst_mat = self.calc_inst_matrix()
         inst_mat = np.around(inst_mat, decimals=5).tolist()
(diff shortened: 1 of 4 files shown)


ziw-liu:
One thing to notice is that these lines used to automatically update the fields in the Acq/Rec tab after loading a calibration:
https://github.com/mehta-lab/recOrder/blob/f7946b914c22de1e7ff3deb1b6d76c07e0590fb3/recOrder/plugin/widget/main_widget.py#L2162-L2169
Any workflow relying on this behavior will be affected.

talonchandler (approved):

talonchandler:
Thanks Ziwen...this is perfect. 

I tested:
- run calibration
- load calibration (with and without microscope params in the calibration file)
- offline mode QLIPP reconstruction (with and without microscope params in the calibration file)
and all worked as expected. 

Thanks for flagging the fact that the parameters will no longer be populated during "Load Calibration". I think the new behavior is correct.

PR #228, opened 2022-10-07
Block `imageio` versions that cause import error

ziw-liu:
Now `napari` installed by `pip` on Windows 10 will not run when one of some specific versions of `imageio` is fetched. This issue is somehow not causing a failure in our test-and-deploy checks.
For releasing 0.2.0 we will be matching [`napari`'s solution](https://github.com/napari/napari/pull/5168).


ziw-liu:
Tested on Automaton (install, start, connect MM, calibration) and Hummingbird (install, start, connect).

talonchandler (approved):

PR #229, opened 2022-10-08
Introduce Black formatting into the code base

ziw-liu:
Redoing #179 as it's stale. This PR will not ruin `git blame` on GitHub, and the same effect can be achieved locally with `git config blame.ignoreRevsFile .git-blame-ignore-revs` in the root directory of this repository.

This PR will start to address #176.


talonchandler (approved):
Very nice! Thanks for the clear instructions and for going through the effort of preserving the git blame info.

I've tested basic functionality on hummingbird (calibration, acquisition, bkg correction) and added some docs, so this is ready to merge from my perspective.

PR #232, opened 2022-10-12
git blame ignore squashed commit hash

talonchandler:
This repo squashes the commits when we merge a PR, so the `.git-blame-ignore-revs` file was ignoring a hash that no longer existed in the log. 

This PR updates now tell git blame to ignore the new squashed commit.


ziw-liu:
Sorry I forgot to change the merging option...

ziw-liu (approved):

PR #233, opened 2022-10-13
Update development guide to install `waveorder` for co-development

ziw-liu:
To address #231.

<del>
This quick attempt may not be a good long-term solution if we plan to release unstable `waveorder` to PyPI in the future. (Then we will have to pin the `waveorder` version manually before each release.)
</del>(Not an appropriate solution.)

*Update: It seems like we can just ignore the 'error' message from `pip install ./waveorder -e ".[dev]"` when we already have `recOrder` installed -- the editable installations still work.*

Alternatively, we can:

In the `[options.extras_require]` section of `setup.cfg`, add a switch for different `waveorder` versions. But this will complicate `pip isntall` on the user side. An example is how [napari handles different Qt backends](https://github.com/napari/napari/blob/c305b4607569b963ca4862172e9ef8a9715b52bc/setup.cfg#L91-L107):
```cfg
[options.extras_require]
pyside2 =
    PySide2>=5.13.2,!=5.15.0 ; python_version != '3.8'
    PySide2>=5.14.2,!=5.15.0 ; python_version == '3.8'
pyside =  # alias for pyside2
    %(pyside2)s
pyqt5 =
    PyQt5>=5.12.3,!=5.15.0
pyqt =  # alias for pyqt5
    %(pyqt5)s
qt =  # alias for pyqt5
    %(pyqt5)s
# all is the full "batteries included" extra.
all =
    scikit-image[data]
    %(pyqt5)s
```

### DIFF
--- a/.github/workflows/plugin_preview.yml
+++ b/.github/workflows/plugin_preview.yml
@@ -16,6 +16,6 @@ jobs:
         uses: actions/checkout@v2
 
       - name: napari hub Preview Page Builder
-        uses: chanzuckerberg/napari-hub-preview-action@v0.1.5
+        uses: chanzuckerberg/napari-hub-preview-action@v0.1
         with:
           hub-ref: main
--- a/docs/development-guide.md
+++ b/docs/development-guide.md
@@ -92,7 +92,7 @@ Although many of `recOrder`'s tests are automated, many features require manual 
 
 We use `QT Creator` for large parts of `recOrder`'s GUI. To modify the GUI, install `QT Creator` from [its website](https://www.qt.io/product/development-tools) or with `brew install --cask qt-creator`
 
-Open `/recOrder/recOrder/plugin/qtdesigner/recOrder_ui.ui` in `QT Creator` and make your changes.
+Open `/recOrder/recOrder/plugin/gui.ui` in `QT Creator` and make your changes. 
 
 Finally, convert the `.ui` to a `.py` file with:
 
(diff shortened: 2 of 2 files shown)


talonchandler:
Just tested on hummingbird...works great. VSCode's "Add Folder To Workspace" function works really well, and lets you handle commits to both repos from within the git tab.

talonchandler (approved):

talonchandler:
@ziw-liu Is there a way to "Go to Definition" of a `waveorder` function from a `recOrder` file? I briefly poked around and I couldn't find obvious VSCode support for this.

ziw-liu:
> Is there a way to "Go to Definition" of a waveorder function from a recOrder file? 

For me "Go to Definition" just works (tested in `waveorder_microscopy` called in `recOrder.compute.qlipp_compute`). Check if you have set up the [import discovering workaround](https://github.com/mehta-lab/recOrder/blob/0be6260d8bb92e4e60e542105471e929a9b7f3a5/docs/development-guide.md?plain=1#L48-L49) described in the docs.

talonchandler:
Great, thanks. 

I have a `recOrder` directory alongside a `waveorder` directory. When I added `recOrder` to my workspace, then`waveorder`, my `waveorder` imports in `recOrder` didn't work until I added:
```
{
    "python.analysis.extraPaths": ["../waveorder"]
}
```
to a newly created `recOrder/.vscode/settings.json`. Now everything's working as expected. 

Is this the best way to handle this? Should I add these instructions to docs?

ziw-liu:
> Is this the best way to handle this? Should I add these instructions to docs?

I don't know if there's a better way since it's more like a `pip` issue than an IDE bug. If a developer followed the current documentation they should be able to perform static analysis just fine.

PR #239, opened 2022-10-19
Test MM Core calls with mocks

ziw-liu:
Our test coverage has been declining. This is an initial attempt to address this problem by using simple mock-up objects to unit-test methods that make calls to Micro-Manger APIs.

On a side note, it may be beneficial if we isolated all outbound calls via `pycromanager` and aggregate them in the `io.core_functions` module.

Starting to address #156.

### DIFF
--- a/recOrder/tests/mmcore_tests/test_core_func.py
+++ b/recOrder/tests/mmcore_tests/test_core_func.py
@@ -1,21 +1,32 @@
-from unittest.mock import MagicMock, Mock
-from numpy.random import randint
+from unittest.mock import MagicMock, Mock, call
+import pytest
+import numpy as np
+from numpy import ndarray
+from typing import Callable
 
 # tested components
 from recOrder.io.core_functions import *
 
 
+# TODO: move these to fixture or generate with Hypothesis
 # dynamic range
 TIFF_I_MAX = 2**16
 # image size
-IMAGE_WIDTH = randint(1, 2**12)
-IMAGE_HEIGHT = randint(1, 2**12)
+IMAGE_WIDTH = np.random.randint(1, 2**12)
+IMAGE_HEIGHT = np.random.randint(1, 2**12)
 PIXEL_COUNT = IMAGE_HEIGHT * IMAGE_WIDTH
 # serialized image from the pycromanager bridge
-SERIAL_IMAGE = randint(0, TIFF_I_MAX, size=(PIXEL_COUNT))
+SERIAL_IMAGE = np.random.randint(0, TIFF_I_MAX, size=(PIXEL_COUNT,))
+# LC device parameters
+# TODO: parameterize this example
+DEVICE_PROPERTY = ("deviceName", "propertyName")
+CONFIG_GROUP = "configGroup"
+CONFIG_NAME = "State0"
+# LC state in native units
+LC_STATE = np.random.rand(1)[0] ** 10
 
 
-def get_mmcore_mock():
+def _get_mmcore_mock():
     """Creates a mock for the `pycromanager.Core` object.
 
     Returns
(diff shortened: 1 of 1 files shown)


recOrder/tests/mmcore_tests/test_core_func.py
```diff
@@ -0,0 +1,212 @@
...
+# tested components
+from recOrder.io.core_functions import *
+
+
+# TODO: move these to fixture or generate with Hypothesis
+# dynamic range
+TIFF_I_MAX = 2**16
+# image size
+IMAGE_WIDTH = np.random.randint(1, 2**12)
+IMAGE_HEIGHT = np.random.randint(1, 2**12)
```
talonchandler, line 16: Cool! This is a really nice place for randomness.

recOrder/tests/mmcore_tests/test_core_func.py
```diff
@@ -0,0 +1,212 @@
...
+PIXEL_COUNT = IMAGE_HEIGHT * IMAGE_WIDTH
+# serialized image from the pycromanager bridge
+SERIAL_IMAGE = np.random.randint(0, TIFF_I_MAX, size=(PIXEL_COUNT,))
```
talonchandler, line 19: Fantastic. I think on future iterations we can have a really simple noise model here. No changes needed now.

recOrder/tests/mmcore_tests/test_core_func.py
```diff
@@ -0,0 +1,212 @@
...
+    mmc = _get_mmcore_mock()
+    image = snap_image(mmc)
+    mmc.snapImage.assert_called_once()
+    mmc.getImage.assert_called_once()
+    assert _is_int(image), image.dtype
+    assert image.shape == (PIXEL_COUNT,), image.shape
+
+
+def test_suspend_live_mm():
+    """Test `recOrder.io.core_functions.suspend_live_mm`."""
```
talonchandler, line 123: I think `mm` -> `sm`, right?

talonchandler (approved):
I think this is an excellent step towards improved test coverage, and I think you've made some really excellent choices here. Great job, Ziwen!

I really like that you're testing the `io.core_functions` first, and I second your idea that we should work towards isolating all of our pycromanager/external calls  into `io.core_functions`. We're likely not far from that goal when we begin full work towards 1.0.0. 

I particularly like:
- that you've made the test image width and height random. This isn't exciting yet, but we've seen all sorts of bugs with the user selecting different ROIs (for example, see today's #236) not to mention challenging background correction bugs. 
- that you've marked much of the initialization code to be moved to a fixture when we're ready. I think it's great that you're taking this step by step. 

I think this is ready to merge since it tests one file cleanly. 

What do you think the next file/functionality to test is? I'd be particularly excited to see the calibration code tested since it will not be changing anytime soon, and we have a long list of setting combinations that we'd like to test. I can help you set up simple simulations of the LCs and cameras that will let us do a simulate a complete calibration.

recOrder/tests/mmcore_tests/test_core_func.py
ziw-liu: Fixed!

ziw-liu:
> I'd be particularly excited to see the calibration code tested since it will not be changing anytime soon, and we have a long list of setting combinations that we'd like to test.

I think this makes a lot of sense. One thing I (intentionally) left out in this PR is the generation of test data. I am almost convinced by [this paper](https://conference.scipy.org/proceedings/scipy2020/zac_hatfield-dodds.html) that we should use [Hypothesis](https://github.com/HypothesisWorks/hypothesis) to do this, and I think we can have a better understanding of what do we need from such a tool once we start to conceive of testing the most physics-heavy part of `recOrder`.

talonchandler:
Just skimmed the paper...looks pretty cool. I think it's definitely worth a try.

recOrder/tests/mmcore_tests/test_core_func.py
```diff
@@ -0,0 +1,212 @@
...
+from numpy import ndarray
+from typing import Callable, Tuple
+
+# tested components
+from recOrder.io.core_functions import *
+
+
+# TODO: move these to fixture or generate with Hypothesis
+# dynamic range
+TIFF_I_MAX = 2**16
```
ieivanov, line 13: I'd recommend also testing with 8-bit and 12-bit images in the future.

ieivanov:
This is great! Thanks for getting it started @ziw-liu.

PR #245, opened 2022-10-27
Turn off autoshutter before bg capture

### LINKED ISSUE #237: [BUG] Illumination is shuttered during background correction
Illumination is shuttered between channels during background correction, which can cause problems if the camera and shutter are not well coordinated. It will be better is the light is turn on at the beginning of collecting background images, left on between channels, and turn off at the end, like we do during data acquisition.

ieivanov:
Fixes #237 

Note that turning off autoshutter does not require an automatic shutter, i.e. this code will run even if the microscope has only a manual shutter


ziw-liu (approved):
Nice!

talonchandler:
Just tested on hummingbird...works like a charm! Thanks @ieivanov

PR #246, opened 2022-10-27
Simplify acquisition settings

talonchandler:
Following our discussion in group meeting, I've made the following simplifying changes to the "Acquisition Settings" panel:

- Instead of two 2D vs. 3D menus birefringence and phase, a single 2D vs. 3D menus
- Renamed "Acq Channel (BF Only)" to "BF Channel"
- "Phase from BF" checkbox aligned with "BF Channel" selection box

Screen Shot 2022-10-27 at 3 01 11 PM


ziw-liu (commented):
Code looks good to me!

Do you think there is a need to document what does 'Acquisition Mode' mean? It was not immediately clear to me from the GUI if the 2D and 3D modes for phase acquisition are capturing different raw data or doing different reconstruction.

ziw-liu (approved):

talonchandler:
Good call...I added a couple tooltips. Thanks @ziw-liu.

PR #248, opened 2022-11-01
90 degree orientation offset checkbox

talonchandler:
Thanks for offering to help finish and test this, Ziwen. 

In this draft PR I've placed the GUI elements...the relevant checkbox is named `cb_orientation_offset`. 

The expected behavior is:
- unchecked by default...no change to to the reconstructions or visualization
- if checked, both the reconstruction and the visualization of the reconstruction will be rotated by 90 degree---I think you'll like need to use apply something like `pi/2 modulo pi` to the existing reconstruction.


talonchandler:
Just tested on hummingbird with Ed and it's working as I expect. Thanks for adding the warning message. 

I can't approve a PR that I created, so can you approve and merge this, Ziwen? Thanks!

ziw-liu:
> can you approve and merge this

Ideally there's some code review since I experienced some confusing behavior during local testing (most likely due to some cache issue with editable install, disappeared after rebooting the computer). Maybe @edyoshikun can take a look if you haven't already?

ziw-liu (approved):

PR #252, opened 2022-11-04
Import `qtpy` instead of `PyQt5` in autogenerated `gui.py`

talonchandler:
This fix requires "manually" changing the import in the autogenerated `gui.py` file, but I've made this as easy as possible by adding a `sed` one-liner to the developer documentation.


ziw-liu:
I just merged #233, introducing a conflict between this and HEAD. @talonchandler Can you take a look and resolve it?

ziw-liu (approved):

PR #256, opened 2022-11-07
Override initial tab to focus on 'LC Calibration'

### LINKED ISSUE #254: On startup, plugin opens "Acquisition / Reconstruction" tab instead of "LC Calibration" tab
Annoyingly, Qt Creator sets the startup tab focus to the last tab you were editing when you save the .ui file, so when I edit the .ui file I sometimes accidentally change the tab focus. 

The current behavior is an example of this problem: on startup the plugin opens to the "Acquisition / Reconstruction" tab instead of the expected "LC Calibration" tab.

@ziw-liu would you mind adding a couple lines to the `main_widget` constructor to fix the tab focus?

ziw-liu:
Fixes #254

### DIFF
--- a/recOrder/plugin/main_widget.py
+++ b/recOrder/plugin/main_widget.py
@@ -50,6 +50,9 @@ class MainWidget(QWidget):
         self.ui = gui.Ui_Form()
         self.ui.setupUi(self)
 
+        # Override inital tab focus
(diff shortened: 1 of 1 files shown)


talonchandler (approved):
Works perfectly! Thanks @ziw-liu.

PR #261, opened 2022-11-08
Remove background debugging print statement

talonchandler:
This PR removes a debugging print statement that was left behind, solving #258. 

@mattersoflight brought up a potential concern that "local_fit" background corrections were being applied to phase reconstructions, but I have reread the code closely and found that this is not the case. Background corrections are only being applied to polarization acquisition.

### DIFF
--- a/recOrder/compute/qlipp_compute.py
+++ b/recOrder/compute/qlipp_compute.py
@@ -16,7 +16,7 @@ def initialize_reconstructor(
     z_step_um=None,
     pad_z=0,
     pixel_size_um=None,
-    bg_correction="local_fit",
+    bg_correction="None",
     n_obj_media=1.0,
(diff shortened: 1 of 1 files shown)


ziw-liu:
I found this behavior confusing too. Also when initializing the reconstructor for phase from BF, maybe we can explicitly pass in `None` for the `bg_correction` keyword argument (which defaults to `"local_fit"`) here:
https://github.com/mehta-lab/recOrder/blob/23f6edd864eafbc16be6b07f9c231f0abe1be178/recOrder/acq/acquisition_workers.py#L230-L247

ziw-liu (approved):

talonchandler:
Agreed. Changing the keyword argument to "None" is much cleaner. Thanks!

PR #270, opened 2022-11-09
Draw overlay for both 2D and 3D birefringence acquisitions

### LINKED ISSUE #263: Overlay is not generated for 3D "Retardance + Orientation" reconstructions
Acquiring "Retardance + Orientation" in 2D will result in an "Overlay" layer, but in 3D mode not overlay is generated. Can we use the same overlay settings and generate a 3D overlay by default?

ziw-liu:
Resolves #263

### DIFF
--- a/recOrder/plugin/main_widget.py
+++ b/recOrder/plugin/main_widget.py
@@ -22,6 +22,7 @@ from waveorder.io.reader import WaveorderReader
 from pathlib import Path, PurePath
 from napari import Viewer
 from napari.utils.notifications import show_warning
+from napari.qt.threading import create_worker
 from numpydoc.docscrape import NumpyDocString
 from packaging import version
 import numpy as np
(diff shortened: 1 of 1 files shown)


ziw-liu:
@talonchandler This naive solution does leave the GUI window unresponsive after reconstruction finished on the Automaton computer. Not sure if it is because of [this](https://github.com/mehta-lab/recOrder/issues/263#issuecomment-1307714044) though.

ziw-liu:
Now the GUI does not freeze, but there's a delay before 3D overlay can be updated. On a local computer with high single-thread CPU performance, a 21*512*512 volume (from demo MM camera) takes < 1s; with Xeon Silver @2.4GHz we typically have on microscope computers, a 21*2048*2048 volume will take 5~10s to be rendered.

talonchandler (approved):
This is awesome! I didn't realize we could spin up qt thread this easily. 

Works extremely well. Thanks @ziw-liu

PR #273, opened 2022-11-14
Read illumination parameters from the loaded calibration instead of GUI during acquisition

### LINKED ISSUE #255: Acquisition reads from "LC Calibration" tab instead of the most recently run/loaded calibration
The following steps illustrate the problem:
- Set "Illumination Scheme" to "5-State"
- "Run Calibration"
- "Acquire Birefringence"
- Set "Illumination Scheme" to "4-State" (without re-running the calibration)
- "Acquire Birefringence" --- fails with a mismatched shape error:
`ValueError(f'Unsupported image data size. Provide image data is of size: {data_dims}.'`

I think `recOrder` needs to keep it's internal calibration state separate from the GUI. In this case, `enter_calib_scheme` is run whenever the GUI changes, which updates `self.calib_scheme`, which is read and used by `acquisition_workers`. 

I might suggest that `QLIPP_Calibration` gets new attributes that are used to keep track of the latest calibration. With these new attributes, the `acquisition_workers` can call `self.calib_window.calib.calib_scheme` instead of the current `self.calib_window.calib_scheme`.

I have similar recommendations for handling `swing` and `wavelength`.

I'm very open to different solutions, too! Thanks @ziw-liu.

ziw-liu:
Fixes #255

### DIFF
--- a/recOrder/acq/acquisition_workers.py
+++ b/recOrder/acq/acquisition_workers.py
@@ -603,7 +603,7 @@ class PolarizationAcquisitionWorker(WorkerBase):
 
         # List the Channels to acquire, if 5-state then append 5th channel
         channels = ["State0", "State1", "State2", "State3"]
-        if self.calib_window.calib_scheme == "5-State":
+        if self.calib.calib_scheme == "5-State":
             channels.append("State4")
 
         self._check_abort()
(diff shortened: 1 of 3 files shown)


ziw-liu:
Test failed on pip network error. Running again.

ziw-liu:
There might still be minor improvements left on the table, but will proceed to merge since the current implementation does fix the issue.

talonchandler (approved):
Good catch! Ready to go from my perspective.

PR #275, opened 2022-11-15
Add ROI size check for the previous phase reconstructor

### LINKED ISSUE #236: Phase acquisitions fail after changing MM ROI
recOrder tries to precompute, save, and reuse OTFs so that it doesn't needlessly recompute identical OTFs during each acquisition. 

This functionality works well when there's a recOrder-side parameter change---the OTFs are recomputed correctly. However, the OTFs are not recomputed when we change the acquired ROI in MM, and the reconstruction will fail. 

Steps to reproduce:
- "Acquire Phase" from recOrder (OTF is computed and saved)
- Change the ROI in MM
- "Acquire Phase" from recOrder (OTF is not recomputed and recOrder errors with mismatched dimensions)

We can work around this bug by restarting recOrder so that the OTFs are computed from scratch.

I don't think this is a high-priority bug, but I think it should be handled correctly for 1.0.0.

ziw-liu:
Fixes #236

### DIFF
--- a/recOrder/acq/acquisition_workers.py
+++ b/recOrder/acq/acquisition_workers.py
@@ -254,7 +254,7 @@ class BFAcquisitionWorker(WorkerBase):
             self._check_abort()
 
             # compute new reconstructor if the old reconstructor properties have been modified
-            if self._reconstructor_changed():
+            if self._reconstructor_changed(stack.shape):
                 logging.debug(
                     "Reconstruction settings changed, updating reconstructor"
                 )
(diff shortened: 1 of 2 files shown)


recOrder/acq/acquisition_workers.py
```diff
@@ -413,60 +414,40 @@ def _reconstructor_changed(self):
...
+                ):
+                    return True
+            elif key == "n_slices":
+                if getattr(self, key) != getattr(old_recon, value):
+                    return True
+            elif key == "ps":
+                if getattr(self.calib_window, key) / float(
+                    self.calib_window.mag
+                ) != getattr(old_recon, value):
+                    return True
```
talonchandler, line 443: Excellent improvement!

talonchandler (approved):
Excellent! Thanks for handling this @ziw-liu. I tested on hummingbird and it's working extremely well. Merging...

PR #279, opened 2022-11-16
Save GUI state with snapped images

### LINKED ISSUE #262: GUI state is not saved with data
@mattersoflight and I discussed `recOrder` workflows and agreed that saving `recOrder`'s GUI state with each acquisition will be useful, even if the format is mostly for human eyes (since we're putting off truly reproducible reconstructions until 1.0). 

Expected behavior: a text file is written into the folder of each acquisition (background folders and "Acquire" folders) containing the GUI settings of the 
- "Calibration Parameters" box
- "Capture Background" box
- "Acquisition Settings" box
- "General Reconstruction Settings" box
- "Phase Reconstruction Settings" box

@ziw-liu do you think I should wait until #255 is fixed to write this? I would like to save the "Calibration Parameters" that correspond to the most recent calibration instead of the GUI.

ziw-liu:
Resolves #262.

### DIFF
--- a/recOrder/plugin/main_widget.py
+++ b/recOrder/plugin/main_widget.py
@@ -34,6 +34,11 @@ import logging
 import textwrap
 import yaml
 
+# type hint/check
+from typing import TYPE_CHECKING
+if TYPE_CHECKING:
+    from _typeshed import StrOrBytesPath
+
 
 class MainWidget(QWidget):
     """
@@ -136,7 +141,9 @@ class MainWidget(QWidget):
         )
 
         # This parameter seems to be wired differently than others...investigate later
-        self.ui.le_recon_wavelength.editingFinished.connect(self.enter_recon_wavelength)
+        self.ui.le_recon_wavelength.editingFinished.connect(
+            self.enter_recon_wavelength
+        )
         self.ui.le_recon_wavelength.setText("532")
         self.enter_recon_wavelength()
 
@@ -2015,8 +2022,14 @@ class MainWidget(QWidget):
         # Start Thread
         self.worker.start()
 
-    @Slot()
-    def _dump_gui_state(self, save_dir):
+    def _dump_gui_state(self, save_dir: StrOrBytesPath):
+        """Collect and save the current GUI settings to a YAML file.
+
+        Parameters
+        ----------
+        save_dir : str | bytes | PathLike[str] | PathLike[bytes]
+            directory to save
+        """
         gui_state = {
             "Run Calibration": {
                 "Swing": self.swing,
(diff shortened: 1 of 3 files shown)


recOrder/plugin/main_widget.py
```diff
@@ -2014,6 +2027,51 @@ def acq_ret_ori_phase(self):
...
+                "Objective NA": self.obj_na,
+                "Condenser NA": self.cond_na,
+                "Camera Pixel Size": self.ps,
+                "RI of Objective Media": self.n_media,
+                "Magnification": self.mag,
+                "Orientation Offset": self.orientation_offset,
+            },
+        }
+        save_path = os.path.join(save_dir, "gui_state.yml")
+        with open(save_path, "w") as f:
```
talonchandler, line 2072: This is great! This is exactly the format I had in mind. 

Do you mind:
1. Adding the "Phase Reconstruction Settings"? I think there are just "Z Padding", "Regularizer", and "Strength"
2. Figuring out why the order changed on hummingbird (see attached)? I'm guessing it's related to `yaml.dump`?
[gui_state.txt](https://github.com/mehta-lab/recOrder/files/10026859/gui_state.txt)

Otherwise, LGTM!

ziw-liu: > Adding the "Phase Reconstruction Settings"

I meant to add those as well... slipped over my mind (again!)

talonchandler:
When I acquired a background I received:
`AttributeError: 'MainWidget' object has no attribute 'phase_regularizer'`
because `phase_regularizer` hadn't been created yet. 

I added a one-line fix calling `enter_phase_denoiser` in the `main_widget` constructor.

talonchandler:
@Soorya19Pradeep @edyoshikun this PR addresses one of the issues we've run into recently: what parameter did you choose in `recOrder` when you acquired and reconstructed an image?

Our solution for 0.3.0 is to save a `gui_state.yml` file with each acquisition. Here's an example of one of these files:
```
Run Calibration:
  Swing: 0.1
  Wavelength: 532
  Illumination Scheme: 4-State
  Calibration Mode: MM-Voltage
  Config Group: Channel
Capture Background:
  Background Folder Name: BG
  Number of Images to Average: 5
Acquisition Settings:
  Z Start: -10.0
  Z End: 10.0
  Z Step: 1.0
  Acquisition Mode: 2D
  BF Channel: BF
General Reconstruction Settings:
  Background Correction: None
  Background Path: C:\Users\LabelFree
  Wavelength: 532
  Objective NA: 1.3
  Condenser NA: 0.5
  Camera Pixel Size: 6.9
  RI of Objective Media: 1.3
  Magnification: 60.0
  Orientation Offset: false
Phase Reconstruction Settings:
  Z Padding: 0
  Regularizer: Tikhonov
  Strength: 0.0001
```
Any comments @edyoshikun @Soorya19Pradeep? Does this solve most of the issues we've encountered?

This solution is a stop-gap on the way to 1.0.0 where we are aiming for the complete reconstruction to be exactly reproducible very quickly. For now, we think it's important to save the GUI state so that you can modify `recOrder` scripts.

recOrder/plugin/main_widget.py
```diff
@@ -234,6 +234,7 @@ def __init__(self, napari_viewer: Viewer):
         self.ui.cb_phase_denoiser.currentIndexChanged[int].connect(
             self.enter_phase_denoiser
         )
+        self.enter_phase_denoiser()
```
ziw-liu, line 237: This will default `MainWidget.phase_regularizer` to `"Tikhonov"`. I think this behavior is fine, yet not sure if we should declare it explicitly.

talonchandler (approved):
Works really well...and I've received positive feedback from in-person conversations. Merging now...

PR #282, opened 2022-11-17
Fix 'Load Calibration' when Micro-Manager is in Live Mode

### LINKED ISSUE #280: "Load calibration" when MM is live gives an error
Clicking "Load calibration" when MM is live errors and gives the attached traceback:

[traceback.txt](https://github.com/mehta-lab/recOrder/files/10026782/traceback.txt)

Easy to work around by turning off "Live Mode". 

Related to #203 and #205.

### LINKED ISSUE #281: "Load calibration" when MM is live gives an error
Clicking "Load calibration" when MM is live errors and gives the attached traceback:

[traceback.txt](https://github.com/mehta-lab/recOrder/files/10026782/traceback.txt)

Easy to work around by turning off "Live Mode". 

Related to mehta-lab/recOrder#203 and mehta-lab/recOrder#205.

ziw-liu:
Fix mehta-lab/waveorder#281


ziw-liu:
Briefly tested on the Hummingbird.

@talonchandler This is a larger change than #335 but I think avoiding MM calls in the main thread as much as possible and actually fixing the calibration code is a cleaner solution.

talonchandler:
I'm a little bit lost here, so please take these questions as naive.  

If the new goal of this PR is to avoid MM calls in the main thread (which is a good goal IMO), then I think these lines in `run_calibration` need to be removed too:
https://github.com/mehta-lab/recOrder/blob/add3280ad368971d63ebfd230ab22fb8e2ca862f/recOrder/plugin/main_widget.py#L1899-L1901

But I tried removing them, and I received an error (traceback to L435 of `Calibration.py` with `set_lc_state`) when I tried running a calibration with MM in live mode. Do we need the new decorator in more places in `Calibration.py`

Also, does the new decorator mean that we not longer need the `suspend_live_sm` contextmanager? They seem to be doing roughly the same thing?

I'm open to:
- handling these issues "correctly" with no MM calls on the main thread and only one way of handling MM live mode. Following this path we can abandon #335 and merge this after more iteration. 
- Or we can merge #335, open a new issue about getting MM calls off of the main thread, and bump these changes to later.

ziw-liu:
> Also, does the new decorator mean that we not longer need the `suspend_live_sm` contextmanager? They seem to be doing roughly the same thing?

Roughly yes, but they serve different scenarios. The decorator will turn off live mode for the whole decorated function, whereas the context manager is a more fine-grained control. Also, the decorator is not flexible enough if the function/method needs to use the (modified) snap manager object, because that would require that they take a fixed argument for the object.

talonchandler:
Sorry @ziw-liu I accidentally modified this branch (I think with unchanged functionality). 

Issue #335 is fixed on the 0.4.0dev branch, and I'm not sure what should happen with this branch

ziw-liu:
Closing as this has become stale, and calibration will undergo a major refactor in v1.0.0.

PR #286, opened 2022-11-22
Restore previous MDA settings after acquisition

### LINKED ISSUE #253: recOrder acquisitions modify the MDA window
recOrder relies on the MDA to acquire phase and birefringence and as a result modifies the MDA settings.

For example, the user may be using the MDA to acquire Cy3 and Cy5 channels. Then if they snap a biref image, recOrder will modify the MDA channels to State0, State1, etc. It will be nice if recOrder remembers the previous MDA settings and changes back to them after finishing the snap.

ziw-liu:
Fix #253

### DIFF
--- a/recOrder/acq/acq_functions.py
+++ b/recOrder/acq/acq_functions.py
@@ -4,6 +4,7 @@ import os
 from waveorder.io.reader import WaveorderReader
 import time
 import glob
+from pycromanager import Studio
 
 
 def generate_acq_settings(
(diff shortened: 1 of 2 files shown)


talonchandler:
I just tested and found that this works as expected **when the MDA window is closed**. 

**Working test**
1) Open the MDA window, set a bunch of custom settings, close the MDA window
2) Run an acquisition in recOrder
3) Reopen the MDA window in MM and see that all of the custom settings were saved. 

**Failing test**
1) Open the MDA window, set a bunch of custom settings, leave the MDA window open
2) Run an acquisition in recOrder and watch the custom settings disappear
3) The MDA custom settings never repopulate.

Maybe you can programmatically close the MDA window before each `recOrder` acquisition then reopen it after?

talonchandler:
I just took a closer look to see if I could fix this...and now I can't reproduce my own bug!

The MDA acquisitions settings seem to be repopulating now whether or not the MDA window is open. 

I will continue my testing session and see if I can figure out what caused the issue.

talonchandler (approved):
I just tested on falcon, and it's working really well. Let's keep an eye out for weird behavior, but this is ready to merge from my perspective. 

I'll let you merge, @ziw-liu, in case you have any thoughts.

ieivanov (approved):
Tested, works as expected.

PR #287, opened 2022-11-24
Populate "BF Channel" list with presets containing keywords

talonchandler:
Before this PR, the "BF Channel" list would be populated with all MM presets, which would lead to confusing defaults like this:
image

After this PR, the "BF Channel" list will only contain MM presets that have any of the following case-insensitive keywords as a substring:
`["bf", "brightfield", "bright", "labelfree", "label-free", "lf", "label"]`

I checked these keywords against the latest MM configs on `CompMicro_MMConfigs` and tested in person on hummingbird and falcon. Feel free to suggest/add other keywords.

On hummingbird, the startup menus now look like this (similar on falcon):
image


ieivanov:
Another good keyword may be `phase`

ieivanov (approved):

PR #288, opened 2022-11-28
Check for non-physical NA arguments

talonchandler:
This PR throws a `ValueError` if the user requests a reconstruction with the illumination or detection NA are greater than the index of refraction of the media: these correspond to non-physical parameters. 

If the user uses, for example, NA_obj = 1.1 and n_obj_media = 1.0, the user would see:

Before this PR:
`RuntimeWarning: invalid value encountered in sqrt
  oblique_factor = ((1 - lambda_in**2 * fr**2) *Pupil_support)**(1/2) / lambda_in`
along with a `nan` reconstruction. 

After this PR:
`ValueError: Parameters are non-physical: NA_obj > n_obj_media` with no reconstruction.

### DIFF
--- a/recOrder/compute/qlipp_compute.py
+++ b/recOrder/compute/qlipp_compute.py
@@ -98,6 +98,11 @@ def initialize_reconstructor(
 
     anisotropy_only = False
 
+    if NA_obj > n_obj_media:
+        raise ValueError("Parameters are non-physical: NA_obj > n_obj_media")
+    if NA_illu > n_obj_media:
+        raise ValueError("Parameters are non-physical: NA_illu > n_obj_media")
+
     if pipeline == "QLIPP" or pipeline == "PhaseFromBF":
 
         if not NA_obj:


ziw-liu (commented):
I couldn't locate the source of the 'before PR' behavior in `recOrder` or `waveorder`. @talonchandler Can you point me to it?

talonchandler:
Yep! Here's the more complete error message:
```
/Users/talon.chandler/waveorder/waveorder/optics.py:310: RuntimeWarning: invalid value encountered in sqrt
  oblique_factor = ((1 - lambda_in**2 * fr**2) *Pupil_support)**(1/2) / lambda_in
/Users/talon.chandler/waveorder/waveorder/optics.py:354: RuntimeWarning: invalid value encountered in sqrt
  oblique_factor = ((1 - lambda_in**2 * fr**2) *Pupil_support)**(1/2) / lambda_in
```

ziw-liu:
OK so it's not an explicitly warning from `waveorder`, but inferred by the interpreter.

ziw-liu (approved):

PR #291, opened 2022-12-01
Set NA defaults

### LINKED ISSUE #290: `NA_obj` and `NA_illu` checks don't pass with default parameters
"Capture Background" tries to reconstruct without specifying the `NA_obj` or `NA_illu`, which throws an error given the new checks introduced in #288. 

```
File ~\recOrder\recOrder\compute\qlipp_compute.py:101, in initialize_reconstructor(pipeline='birefringence', image_dim=(540, 750), wavelength_nm=532, swing=0.1, calibration_scheme='4-State', NA_obj=None, NA_illu=None, mag=None, n_slices=None, z_step_um=None, in_focus_slice=None, pad_z=0, pixel_size_um=None, bg_correction='None', n_obj_media=1.0, mode='3D', use_gpu=False, gpu_id=0)
     26 """
     27 Initialize the QLIPP reconstructor for downstream tasks. See tags next to parameters
     28 for which parameters are needed for each pipeline
   (...)
     96 
     97 """
     99 anisotropy_only = False
--> 101 if NA_obj > n_obj_media:
        NA_obj = None
        n_obj_media = 1.0
    102     raise ValueError("Parameters are non-physical: NA_obj > n_obj_media")
    103 if NA_illu > n_obj_media:

TypeError: '>' not supported between instances of 'NoneType' and 'float'
```

talonchandler:
Fixes #290.

FWIW I think future integration tests can catch bugs like this.


recOrder/compute/qlipp_compute.py
```diff
@@ -9,8 +9,8 @@ def initialize_reconstructor(
     wavelength_nm=None,
     swing=None,
     calibration_scheme=None,
-    NA_obj=None,
-    NA_illu=None,
+    NA_obj=1.0,
+    NA_illu=1.0,
```
ziw-liu, line 13: Why are these values set to 1.0/1.0 while the defaults in the reconstruction tab are set to 1.3/0.5?

talonchandler: Good point, updated.

ziw-liu (approved):

PR #294, opened 2022-12-10
Disable "Phase From BF" button when no BF channel is found - add tooltip

### LINKED ISSUE #293: [BUG] Bug setting BF channel
If a channel named `BF` (or similar) is not found, the recOrder GUI lists all channels (including things such as `0`, `1`, and `NewPreset`) in the `BF Channel` dropdown menu. I think that's OK, but it may be better if we instead disable this option and ask the user to create a BF channel.

Confider what may happen if the `BF` channel is not in the config group specified in `Config Group` under `LC Calibration`. When playing around I got an error saying `Config group "LF Channel" or its preset "0" does not exist`.

talonchandler:
Fixes #293. 

When no brightfield preset is found in the MM config, `recOrder` will:

**Before this PR:** show a complete list of presets available and let the user acquire a brightfield stack from any of those presets. 
**After this PR:** disable the "Phase From BF" button, disable the "BF Channel" dropdown box (now empty). If the user mouses over either the disabled button or the disabled dropdown, the tooltip in the following screenshot will appear. 
Screen Shot 2022-12-09 at 4 08 57 PM

Note: the user can acquire "Retardance + Orientation" or "Retardance + Orientation + Phase" when no brightfield preset is found because those acquisitions do not require a brightfield channel. 

Edit: extra comma between "the" and "following" in screenshot above removed

### DIFF
--- a/recOrder/plugin/main_widget.py
+++ b/recOrder/plugin/main_widget.py
@@ -207,7 +207,7 @@ class MainWidget(QWidget):
         ]
         msg = (
             f"No brightfield channel found. If you would like to acquire phase from brightfield,"
-            " please restart recOrder after adding a new channel to MicroManager with one of the,"
+            " please restart recOrder after adding a new channel to MicroManager with one of the"
             " following case-insensitive keywords: "
         ) + ", ".join(self.bf_keywords)
         self.no_bf_msg = "\n".join(textwrap.wrap(msg, width=70))
@@ -2123,8 +2123,12 @@ class MainWidget(QWidget):
         }
         # TV-specific parameters
         if self.phase_regularizer == "TV":
-            gui_state["Phase Reconstruction Settings"]["Rho"] = float(self.ui.le_rho.text())
-            gui_state["Phase Reconstruction Settings"]["Iterations"] = int(self.ui.le_itr.text())
+            gui_state["Phase Reconstruction Settings"]["Rho"] = float(
+                self.ui.le_rho.text()
+            )
+            gui_state["Phase Reconstruction Settings"]["Iterations"] = int(
+                self.ui.le_itr.text()
+            )
         # save in YAML
         save_path = os.path.join(save_dir, "gui_state.yml")
         with open(save_path, "w") as f:
--- a/recOrder/plugin/main_widget.py
+++ b/recOrder/plugin/main_widget.py
@@ -190,6 +190,7 @@ class MainWidget(QWidget):
         self.ui.qbutton_acq_phase_from_bf.clicked[bool].connect(
             self.acq_phase_from_bf
         )
+        self.bf_channel_found = False
         self.ui.qbutton_acq_ret_ori_phase.clicked[bool].connect(
             self.acq_ret_ori_phase
         )
@@ -928,7 +929,8 @@ class MainWidget(QWidget):
                     [keyword.lower() in ch.lower() for keyword in bf_keywords]
                 ):
                     self.ui.cb_acq_channel.addItem(ch)
-
+                    bf_channel_found = True
+                    
         if not config_group_found:
             msg = (
                 f"No config group contains channels {self.calib_channels}. "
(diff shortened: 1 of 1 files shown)


recOrder/plugin/main_widget.py
```diff
@@ -190,6 +194,24 @@ def __init__(self, napari_viewer: Viewer):
...
+            "brightfield",
+            "bright",
+            "labelfree",
+            "label-free",
+            "lf",
+            "label",
+            "phase",
+            "ph",
+        ]
+        msg = (
```
ziw-liu, line 208: Why do we initialize this by joining the empty string with the message? AFAIK they are all run in the same `__init__()` method.

recOrder/plugin/main_widget.py
```diff
@@ -190,6 +194,24 @@ def __init__(self, napari_viewer: Viewer):
         self.ui.qbutton_acq_phase_from_bf.clicked[bool].connect(
             self.acq_phase_from_bf
         )
+        self.bf_keywords = [
```
ziw-liu, line 197: This can also be a class attribute if it's never modified in instances.

recOrder/plugin/main_widget.py
talonchandler: I want to use the keywords elsewhere, and I want to append these to the message. Is there an easier way you'd suggest?

recOrder/plugin/main_widget.py
```diff
@@ -598,6 +620,13 @@ def _set_buttons_enabled(self, val):
                 )
                 action_button.setStyleSheet("border: 1px solid rgb(65,72,81);")
 
+        if not self.bf_channel_found:
+            self.ui.qbutton_acq_phase_from_bf.setEnabled(False)
+            self.ui.qbutton_acq_phase_from_bf.setStyleSheet(
+                "border: 1px solid rgb(65,72,81);"
```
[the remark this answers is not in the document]
talonchandler: Good call, will do.

recOrder/plugin/main_widget.py
talonchandler: Can you help me understand when/why I might use class attributes? In this case we'll only have a single instance of this class, right?

(This is a question almost entirely from ignorance...I haven't used class attributes much).

ziw-liu: In this case it does not really matter since nothing is actually private in Python. Here it'll be more like a visual discouragement against modifying what's intended to be a constant. Feel free to leave it this way!

recOrder/plugin/main_widget.py
ziw-liu: Sorry I meant that you don't need this line:
https://github.com/mehta-lab/recOrder/blob/9bfce2833a3f17fc55f0891dbcc302ad1a89b983/recOrder/plugin/main_widget.py#L72
since it's equivalent if you just declare it here:
https://github.com/mehta-lab/recOrder/blob/9bfce2833a3f17fc55f0891dbcc302ad1a89b983/recOrder/plugin/main_widget.py#L213

talonchandler: Aha! Fixed, thank you.

recOrder/plugin/main_widget.py
talonchandler: Thanks! I understand what's going on now.

recOrder/plugin/main_widget.py
```diff
@@ -56,6 +56,26 @@ class MainWidget(QWidget):
...
+        "brightfield",
+        "bright",
+        "labelfree",
+        "label-free",
+        "lf",
+        "label",
+        "phase",
+        "ph",
+    ]
+    msg = (
```
ziw-liu, line 72: Please consider using a more descriptive name for this variable since it's now a class attribute that can be accessed through `MainWidget.msg`.

ziw-liu: I might be missing some constraints here but this should work so that `msg` is not in the namespace:
```py
# delete the `msg = ...` line
no_bf_msg = textwrap.wrap('No ... keywords: '.join(bf_keywords), width=70)
```

talonchandler: Thanks for pointing this out...still learning the ropes of these class attributes. Fixed.

recOrder/plugin/main_widget.py
```diff
@@ -56,6 +56,28 @@ class MainWidget(QWidget):
...
+        "lf",
+        "label",
+        "phase",
+        "ph",
+    ]
+    no_bf_msg = "\n".join(
+        textwrap.wrap(
+            f"No brightfield channel found. If you would like to acquire phase from brightfield,"
+            " please restart recOrder after adding a new channel to MicroManager with one of the"
+            " following case-insensitive keywords: ",
```
ziw-liu, line 76: Missing `bf_keywords`?

talonchandler: Jeez...my bad. Thanks Ziwen.

ziw-liu (approved):

PR #295, opened 2022-12-12
fix BF channel index in 3D-pol-to-birefringence example

ieivanov:


talonchandler (approved):
Great catch! Thank you 👍

PR #297, opened 2022-12-15
`recorder info` and `recorder view` command-line utilities

talonchandler:
I have observed several gaps in our workflows when we're examining datasets:
- What's the approximate shape of this dataset? How many positions? What channels? Do I have to open it in `WaveorderReader` or MM to find out?
- I have a large dataset that I want to take quick look at...what's the best way to open just a few positions?
- By default `recOrder`'s readers opens each position as a layer, but sometimes I want the channels as layers. What's the best way to do this?

**To address these gaps, I've extended `recOrder`'s CLI with two small utilities: `info` and `view`.**

Typing `recOrder` or `recorder` at the command line will print a short help message with all of the subcommands (currently `convert`, `info`, and `view`). 

`recorder info <your/tiff/or/zarr/folder>` will print a message like:
```
Reading file:	 2022_08_04_recOrder_pytest_20x_04NA/2T_3P_16Z_128Y_256X_Kazansky_1/
Positions:	 3
Time points:	 2
Channels:	 4
(Z, Y, X):	 (16, 128, 256)
Channel names:	 ['State0', 'State1', 'State2', 'State3']
Z step (um):	 2.0
```
`recorder view <your/folder>` will open a new napari window and show your dataset, as if you dragged and dropped the folder in. 

`recorder view <your/folder> -p 0 -p 2` will only open the zeroth and second positions with each position set to a different layer. 

`recorder view <your/folder> -p 0 -p 2 -l c` (`-l c` is short for `--layers channel`, which will also work) will open the zeroth and second positions with the channels as layers. 

Opening datasets this way has an additional benefit: we're not opening with a napari `reader`, so we can set the axis slider labels appropriately

Screen Shot 2022-12-14 at 4 56 06 PM

**Changes:** `recOrder.convert` is now `recOrder convert` with the same options as before. This change follows the `click` library's (and most other command-line tools' IME) conventions. 

**Testing status:** I have opened about 25 different datasets acquired with `recOrder`, MM-MDAs, `pycromanager`, and saved from the `WaveorderWriter` and received good results---owing mostly to the good coverage of `WaveorderReader` across our main dataset types. Please report any bugs you spot. 

**To test these utilities, you will need to change to this branch and rerun `pip install -e <path/to/recOrder>`**


recOrder/scripts/cli.py
```diff
@@ -1,40 +1,106 @@
...
 
-    To use recOrder\'s napari plugin, use \033[96mnapari -w recOrder-napari\033[0m
+    napari.run()
 
-    To convert MicroManager .tif files to ome-zarr data format use \033[96mrecOrder.convert\033[0m
 
-    Thank you for using recOrder.
-    """
-    print(help.__doc__)
+cli.add_command(view)
```
ziw-liu, line 100: In the same file, this extra line is equivalent to using `@cli.command()` instead of 
https://github.com/mehta-lab/recOrder/blob/7534ec04ca0b5324abd45e9f734ffd2a212c5730/recOrder/scripts/cli.py#L21

recOrder/scripts/cli.py
```diff
@@ -1,40 +1,106 @@
 import click
+import napari
+import numpy as np
 import sys
```
ziw-liu, line 4: Unused import?

recOrder/scripts/cli.py
```diff
@@ -1,40 +1,106 @@
...
-import os
+
+# Workaround waveorder #97
+# Create napari Viewer before other imports
+v = napari.Viewer()
+v.close()
```
ziw-liu, line 9: Will these lines take the first CLI command ~10s to respond since a QApp has to be spun up and torn down?

recOrder/scripts/cli.py
talonchandler: Nice! Much improved.

ziw-liu (approved):

PR #298, opened 2022-12-15
Example script saves to a single zarr

### LINKED ISSUE #296: `examples/3D-pol-to-birefringence.py` should write phase and birefringence results to a single zarr
Instead of writing to two different zarrs, this example script should write to a single zarr.

talonchandler:
Fixes #296


examples/3D-pol-to-birefringence.py
```diff
@@ -78,24 +78,15 @@
 ## Save to zarr
 # Save birefringence
 writer = WaveorderWriter("./output")
-writer.create_zarr_root("birefringence_" + timestamp)
+writer.create_zarr_root("reconstructions_" + timestamp)
```
ieivanov, line 81: Should we standardize what this zarr store is called?

examples/3D-pol-to-birefringence.py
```diff
@@ -78,24 +78,15 @@
...
 writer.init_array(
     position=0,
-    data_shape=(1, 4, Z, Y, X),
+    data_shape=(1, 5, Z, Y, X),
     chunk_size=(1, 1, 1, Y, X),
-    chan_names=["Retardance", "Orientation", "BF", "Pol"],
+    chan_names=["Retardance", "Orientation", "BF", "Pol", "Phase"],
```
ieivanov, line 86: I think we should rename `Pol` to `DoP` for degree of polarization

examples/3D-pol-to-birefringence.py
talonchandler: Good idea...I just changed all of the scripts to save to `./output/reconstructsion_<timestamp>`. What do you think of that pattern?

examples/3D-pol-to-birefringence.py
talonchandler: Thanks! 

I also changed the name of "BF" to "S0" since that's more description IMO.

talonchandler:
> This PR combines changes to the example scripts with changes to the CLI. I haven't tested the CLI changes yet.

My fault! Fixed now.

examples/3D-pol-to-birefringence.py
ieivanov: Previously we used something called `computed brightfield` for this quantity. I think it's not immediately intuitive that S0 equated to brightfield. I personally think calling it `BF` is good enough for the purpose of a channel name

examples/3D-pol-to-birefringence.py
ieivanov: I'm not sure. I wish it was a bit more descriptive. Maybe something like `recOrder_recon.zarr` to differentiate it from other data the user may put in that directory e.g. `fluor_decon.zarr`?  I'm still not sure if having two zarr stores in the same folder is allowed by ome standards

ziw-liu: > I'm still not sure if having two zarr stores in the same folder is allowed by ome standards

A `*.zarr` store on a local file system is just a directory/folder that contains the Zarr metadata. The Zarr and OME-NGFF specifications allows for any number of `*.zarr/`s in the same directory. Practically you can have as many `*.zarr/` folders in the same directory as the FS can handle.

ieivanov: Perfect, thanks Ziwen. Would Zarr viewers know how to handle multiple .zarr stores? What would happen if I have the following file structure and I drag the `dataset` folder in a viewer like napari?
```
dataset
   FOV1
      data1.zarr
      data2.zarr
   FOV2
      data1.zarr
      data2.zarr
```

ziw-liu (approved):

examples/3D-pol-to-birefringence.py
talonchandler: >> I think it's not immediately intuitive that S0 equated to brightfield

Do we want users to equate S0 to brightfield? I think some distinction between S0 and brightfield is valuable since S0 depends on details of the LCs, calibration, reconstruction, etc.  Maybe `BF-S0` or `BF-pol` or `BF-computed` are okay compromises?

I'm not extremely attached to the distinction, so I'll yield to your final call @ieivanov.

ieivanov: Just returning to this discussion. I think something like `BF - computed` should do

ieivanov (approved):

Issue #300, opened 2022-12-20
`pycromanager==0.19.2` depends on `numpy < 1.24.0`


talonchandler:
Yesterday's `numpy` 1.24.0 release deprecates `np.bool` which `pycromanager` depends on, which is breaking (fresh installations of) `recOrder` including our tests. 

I think we can:
1. Pin to `numpy==1.23.0` and/or
2. Wait for the next release of pycromanager that's compatible with `numpy==1.24.0` (see [this PR](https://github.com/micro-manager/pycro-manager/pull/475) I just opened) and upgrade our pycromanager dependency. 

What do you think @ziw-liu? I'm leaning towards 1 now, then 2 later.

ziw-liu:
Thanks for the PR to upstream!
It makes sense to pin NumPy before PM's next release with the fix.

ziw-liu:
https://github.com/micro-manager/pycro-manager/pull/475 has been merged. However, pycro-manager with this new patch won't be installed with recorder since we have pinned its version:

https://github.com/mehta-lab/recOrder/blob/655a7643f0b9cba3dc51ccb932cf3abc8db38037/setup.cfg#L41

@talonchandler @ieivanov What should we do to confidently bump PM dependency to [latest](https://github.com/micro-manager/pycro-manager/blob/main/pycromanager/_version.py)?

talonchandler:
I think manually testing the MM connection and acquisitions on hummingbird will be enough to upgrade the PM dependency confidently. We might need to upgrade MM as well, but I don't think we'll know until we test. 

I'm not in any rush to upgrade. I think it can wait until after 0.3.0.

talonchandler:
@ziw-liu I just realized that this issue affects new installations of `recOrder==0.2.0`. Would you mind releasing 0.2.1 with `numpy==0.23.5`? 

In the meantime, an easy workaround is `pip install recOrder-napari` followed by `pip install numpy==0.23.5`.

ieivanov:
Now that we have automated testing, pycromanager should be a lot more reliable. We can more confidently upgrade and add tests for features we depend on

ziw-liu:
> @ziw-liu I just realized that this issue affects new installations of `recOrder==0.2.0`. Would you mind releasing a patched version `0.2.1` with pinned `numpy==0.23.5`?

[...]

Issue #172, opened 2022-09-15
`QObject::connect` warning during calibration


talonchandler:
When I start a calibration I receive the following warning:

```
WARNING: QObject::connect: Cannot queue arguments of type 'QTextBlock'
(Make sure 'QTextBlock' is registered using qRegisterMetaType().)
11:01:34 WARNING QObject::connect: Cannot queue arguments of type 'QTextBlock'
(Make sure 'QTextBlock' is registered using qRegisterMetaType().)
QObject::connect: Cannot queue arguments of type 'QTextCursor'
(Make sure 'QTextCursor' is registered using qRegisterMetaType().)
```
I've traced this warning to `_disable_buttons` in `main_widget.py`, but I'm not sure how to fix it. 

@ziw-liu would you mind taking a quick look? It doesn't seem to affect functionality, but it clutters up the terminal...suppressing would be fine.

ziw-liu:
It seems to be occurring when the `CalibrationWorker` calls `logging.info()` [for the first time](https://github.com/mehta-lab/recOrder/blob/418ca4157380c1022d698d280a8eea99fe4694cf/recOrder/plugin/workers/calibration_workers.py#L73). This is somewhat [a common issue for Qt](https://stackoverflow.com/questions/2104779/qobject-qplaintextedit-multithreading-issues) when a worker tries to access something on the other thread. In our case these maybe a log message generated on the `CalibrationWorker` thread and the main widget thread that handles log display.

ziw-liu:
Also I can't fully appreciate having duplicate `logging` messages at different levels, such as:
```py
logging.info('Calculating Black Level ...')
logging.debug('Calculating Black Level ...')
```
I would guess this was a compromise made to display only a certain level of logs in the GUI.

talonchandler:
Aha! I was *very* confused about how stepping through `_disable_buttons` was causing this. You're totally right that threading steered me in the wrong direction. 

I agree that logging could be cleaner. My impression of `logging`'s behavior is that `INFO` is a higher level than `DEBUG`, so `logging.info` should be sufficient, right? See screenshot from here https://docs.python.org/3/library/logging.html

<img width="319" alt="Screen Shot 2022-09-15 at 4 37 12 PM" src="https://user-images.githubusercontent.com/9554101/190526726-2942b094-9b69-4cf3-9941-7ed679feccc6.png">

ziw-liu:
[...]

PR #307, opened 2023-01-16
Updated Pycromanager to 0.24.1.

### LINKED ISSUE #274: `pycromanager==0.19.2` depends on `numpy < 1.24.0`
Yesterday's `numpy` 1.24.0 release deprecates `np.bool` which `pycromanager` depends on, which is breaking (fresh installations of) `recOrder` including our tests. 

I think we can:
1. Pin to `numpy==1.23.0` and/or
2. Wait for the next release of pycromanager that's compatible with `numpy==1.24.0` (see [this PR](https://github.com/micro-manager/pycro-manager/pull/475) I just opened) and upgrade our pycromanager dependency. 

What do you think @ziw-liu? I'm leaning towards 1 now, then 2 later.

nicost:
Connects to MM and disconnects, but fails to reconnect.  That problem seems to be related to ui code I do not understand.

### DIFF
--- a/recOrder/plugin/main_widget.py
+++ b/recOrder/plugin/main_widget.py
@@ -2,7 +2,7 @@
 from __future__ import annotations
 
 from recOrder.calib.Calibration import QLIPP_Calibration, LC_DEVICE_NAME
-from pycromanager import Bridge
+from pycromanager import Core, Studio, zmq_bridge
 from qtpy.QtCore import Slot, Signal, Qt
 from qtpy.QtWidgets import QWidget, QFileDialog, QSizePolicy, QSlider
 from qtpy.QtGui import QPixmap, QColor
@@ -883,9 +883,14 @@ class MainWidget(QWidget):
         RECOMMENDED_MM = "20220920"
         ZMQ_TARGET_VERSION = "4.2.0"
         try:
-            self.bridge = Bridge(convert_camel_case=False)
-            self.mmc = self.bridge.get_core()
-            self.mm = self.bridge.get_studio()
+            self.mmc = Core(convert_camel_case=False)
+            # Check it works
+            self.mmc.getAvailableConfigGroups()
+            self.mm = Studio(convert_camel_case=False)
+            # Order is important: If the bridge is created before Core, Core will not work
+            self.bridge = zmq_bridge._bridge._Bridge()
+            # Happy to use other logging, but we need something
+            print("Established ZMQ Bridge and found Core and Studio")
         except:
             print(
                 (
@@ -898,6 +903,8 @@ class MainWidget(QWidget):
             raise EnvironmentError
 
         # Warn the user if there is a MicroManager/ZMQ version mismatch
+        # NS: Not quite sure what this is good for, we already know the Core works
+        # This code uses undocumented PycroManager features, so may well break in the future
         self.bridge._main_socket.send({"command": "connect", "debug": False})
         reply_json = self.bridge._main_socket.receive(timeout=500)
         zmq_mm_version = reply_json["version"]
@@ -915,6 +922,8 @@ class MainWidget(QWidget):
                 )
             )
 
+        print("Confirmed correct ZMQ bridge----")
+
         # Find config group containing calibration channels
         # calib_channels is typically ['State0', 'State1', 'State2', ...]
         # config_list may be something line ['GFP', 'RFP', 'State0', 'State1', 'State2', ...]
@@ -925,6 +934,7 @@ class MainWidget(QWidget):
         # self.ui.cb_config_group.clear()    # This triggers the enter config we will clear when switching off
         groups = self.mmc.getAvailableConfigGroups()
         config_group_found = False
+        print("Now checking configs")
         for i in range(groups.size()):
             group = groups.get(i)
             configs = self.mmc.getAvailableConfigs(group)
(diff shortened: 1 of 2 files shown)


ziw-liu:
@nicost Thanks!
This can also close #300.

recOrder/plugin/main_widget.py
```diff
@@ -955,7 +965,9 @@ def connect_to_mm(self):
                     self.ui.cb_acq_channel.addItem(ch)
                     self.bf_channel_found = True
 
+        print("Checked configs")
         if not config_group_found:
+            print("Config group not found")
```
ziw-liu, line 970: This message seems to be included in the error message:
https://github.com/mehta-lab/recOrder/blob/7e10ae59a1ed28cf28052b97404a2c0eba4369ff/recOrder/plugin/main_widget.py#L971-L974

recOrder/plugin/main_widget.py
```diff
@@ -990,6 +1003,8 @@ def connect_to_mm(self):
             self.calib_mode = "DAC"
             self.ui.cb_calib_mode.setCurrentIndex(2)
 
+        print("Finished connecting")
```
ziw-liu, line 1006: The `recOrder` plugin widget provides a visual hint of whether the connecting was successful. We may not need this line.

ziw-liu:
Code looks good! Will approve once we have tested on our systems.

recOrder/plugin/main_widget.py
nicost: print statements are all because I saw no other (easy) way to get debug output.  What is the logging mechanism used in recOrder?  I saw that napari has something to output plugin logs, but nothing appeared there.  I would greatly encourage some kind of logging to facilitate future trouble shooting.

ziw-liu: We do have logging with `logging.debug()` etc. in `recOrder`. Unfortunately we do also inherit print statements, and this has been a problem since we are developing a multi-threaded GUI app. See https://github.com/mehta-lab/recOrder/issues/172 for more context and our plan to unify logging with the napari notification system.

edyoshikun:
This seems to be an issue in pycromanager's new implementation of the Core and Studio. The arguments are ignored so for example setting `camel_case= False` does not actually work, so I assume there is something else here that breaks even with pycromanger 0.25.0 updates.

mattersoflight:
@edyoshikun did you figure out a fix? We can now jump to the new version of pycro-manager in preparation of 0.3.0 release.

edyoshikun (commented):
@mattersoflight I just pulled this PR into falcon and had no issues. I wasn't able to reproduce Nico's disconnecting/connecting issue. I tested on micromanager 2.0 092022 and pycromanager 0.24.1 which successfully connected and disconnected from the GUI. I can test pycromanager 0.24.1 on Falcon in next experiments. 

Additionally, most of our calls in recOrder use snake case instead of camel case, so the code is not affected.

ziw-liu:
> Additionally, most of our calls in recOrder use snake case instead of camel case, so the code is not affected.

We do have extensive use of camelCase, for example most of the [recOrder.io.core_functions](https://github.com/mehta-lab/recOrder/blob/d62bdf925f907b8c8620d7d788195bbb336484c5/recOrder/io/core_functions.py).

edyoshikun:
Yes, you are right. I got confused because we use snake case for recOrder functions but camelcase for pycromanager functions. I also realized that restarting my kernel fixed things just like this [issue](https://github.com/micro-manager/pycro-manager/issues/435). 
I just tested the `convert_camel_case` flag and works well. 

```python
from pycromanager import Core
# Case 1
mmc = Core(convert_camel_case=True)
print(mmc.get_available_config_groups())
```
This was a mistake on my part. I have tested with pycromanager `0.24.1` and `0.25.1` and both work. 

One thing to note is that reconnection fails when one closes micromanager, leaves the napari /recOrder window, and then attempts to reconnect(connect to the new mmc instance). This probably deserves it's own separate issue.

edyoshikun:
Updating to `0.24.1` solves the issue of reconnection if the user closes micromanager, leaves the napari/recorder plugin open while still connected, and attempts to reconnect to new instance of micromanager. Previously, I used to see `Failing to Conect` flag and color box change to yellow.

ziw-liu:
@edyoshikun Thanks for looking into this. What's the MM version you used? We should change documentation accordingly if the new PM requires a different version.

edyoshikun:
I used MM2- 09202022 @ziw-liu

ieivanov:
I do support upgrading to a more recent `pycromanager` version. Should we keep the print statements in this code or implement a better logging method?

nicost:
Feel free to take the print statements out or change them to something else.  They were the quickest way for me to get to the source of the problem.  Documenting logging would be a good idea to (I did not find anything about it).

ziw-liu:
> Documenting logging would be a good idea to (I did not find anything about it).

The documentation is [deep into the page](https://github.com/mehta-lab/recOrder/blob/main/docs/napari-plugin-guide.md#advanced-tab) and has an obscure name. We do have plans to rearrange the UI so that logs is easier to access, and that should be realized with the 1.0.0 release.

ziw-liu (approved):
Tested on the Hummingbird and works well.

PR #313, opened 2023-02-07
Update setup.cfg for macos_arm64 specific dependencies

### LINKED ISSUE #193: Native support for Apple Silicon
**Problem**
`recOrder` does not currently support Apple Silicon (M1, M2, etc.) platforms natively due to the incompatibility between `pip` and PyQt5 (PyQt5 does not have a wheel for `arm_64` on PyPI). Native installation would be beneficial for both performance and potential GPU-accelerated processing/ML with a Metal backend.

**Proposed solution**
Closely track [napari's progress on supporting Qt6](https://github.com/napari/napari/issues/3880), since PyQt6/PySide6 have wheels available for the Apple Silicon platform on PyPI. And we can upgrade `recOrder` accordingly when napari makes the move. (They already have initial support merged into the main branch).
This may evolve into a broader set of [1.0.0](https://github.com/mehta-lab/recOrder/milestone/3) issues if napari starts supporting Qt 6 in a stable release soon enough.

**Alternatives**
- Offer a conda environment file that installs PyQt5 from `conda-forge`. This would require the removal of explicit dependency on `napari` in `setup.cfg`. #187

mattersoflight:
make napari and pyqt dependencies platform specific. Fixes #193.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -44,7 +44,10 @@ install_requires =
 	colorspacious>=1.1.2
 	pyqtgraph>=0.12.3
 	napari-ome-zarr>=0.3.2 # drag and drop convenience
-	napari[all] # uses pip-installed PyQt5 by default
+	napari[all];platform_system=='Windows' # uses pip-installed PyQt5 by default
+	napari[all];platform_system=='Linux' # uses pip-installed PyQt5 by default
+	napari;platform_system=='Darwin' # use PyQt6 from conda-forge
+	PyQt6;platform_system=='Darwin' # use PyQt6 from pip
 	importlib-metadata
 	numpy==1.23.5 # TODO: remove after https://github.com/micro-manager/pycro-manager/pull/475
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -44,10 +44,9 @@ install_requires =
 	colorspacious>=1.1.2
 	pyqtgraph>=0.12.3
 	napari-ome-zarr>=0.3.2 # drag and drop convenience
-	napari[all];platform_system=='Windows' # uses pip-installed PyQt5 by default
-	napari[all];platform_system=='Linux' # uses pip-installed PyQt5 by default
-	napari;platform_system=='Darwin' # use PyQt6 from conda-forge
-	PyQt6;platform_system=='Darwin' # use PyQt6 from pip
+	napari[all];'arm64' not in platform_machine # with Qt5 and skimage
+	napari;'arm64' in platform_machine # without Qt and skimage
+	PyQt6;'arm64' in platform_machine
 	importlib-metadata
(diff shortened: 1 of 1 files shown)


setup.cfg
```diff
@@ -44,7 +44,10 @@ install_requires =
 	colorspacious>=1.1.2
 	pyqtgraph>=0.12.3
 	napari-ome-zarr>=0.3.2 # drag and drop convenience
-	napari[all] # uses pip-installed PyQt5 by default
+	napari[all];platform_system=='Windows' # uses pip-installed PyQt5 by default
+	napari[all];platform_system=='Linux' # uses pip-installed PyQt5 by default
+	napari;platform_system=='Darwin' # use PyQt6 from conda-forge
```
ziw-liu, line 49: How would we distinguish Intel-based and M-series based MacOS in this way?

ziw-liu: The `platform_machine` keyword might help. Ref: https://peps.python.org/pep-0508/#environment-markers

E.g. on my M1 mac:

```python_repl
>>> import platform
>>> platform.machine()
'arm64'
```

mattersoflight: It does make sense to use `platform_machine` keyword. 
Quoting from the ref: "The marker language is inspired by Python itself, chosen for the ability to safely evaluate it without running arbitrary code that could become a security vulnerability. " 

Can you try, `napari;platform_system=='Darwin' and platform_machine=='arm64'` and see if that works?

ziw-liu: I tried with:

```config
	napari[all];'arm64' not in platform_machine # with Qt5 and skimage
	napari;'arm64' in platform_machine # without Qt and skimage
	PyQt6;'arm64' in platform_machine
```

And it seems to work on an M1 mac. Still need to test with an Intel mac to confirm (because I'm not sure which arch are the GH test runners using).

image

ziw-liu:
@mattersoflight Do we need this change to be back-ported to 0.2.x (with a v0.2.2 release)?

setup.cfg
talonchandler: At @ziw-liu's request, I just tested an installation on my intel mac---everything seems to be working well (see attached screenshot)
Screen Shot 2023-02-07 at 3 35 02 PM

I also tested on my biohub M1 mac (but with Rosetta I think?). Again, everything seems to be working. 
Screen Shot 2023-02-07 at 3 43 20 PM

Thanks for digging up the `platform_system` keywords @mattersoflight !

ziw-liu: @talonchandler Thanks for help test this!

ziw-liu (approved):
Seems to work on test runners with Windows and Linux too.

mattersoflight:
thanks for quick iteration. Merging. 
@ziw-liu I don't see the need to backport this, because the instructions with 0.2 work as is.

PR #317, opened 2023-02-08
Move imports into cli functions

### LINKED ISSUE #315: `recorder info` fails when used without a screen
The cli needs this workaround for the `view` command to work
https://github.com/mehta-lab/recOrder/blob/d62bdf925f907b8c8620d7d788195bbb336484c5/recOrder/scripts/cli.py#L5-L10

But this workaround means that `recorder info` and `recorder convert` requires a screen (i.e. we can't use ssh). For example:

```
>> recorder info
WARNING: could not connect to display
WARNING: Could not load the Qt platform plugin "xcb" in "" even though it was found.
WARNING: This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Available platform plugins are: eglfs, linuxfb, minimal, minimalegl, offscreen, vnc, wayland-egl, wayland, wayland-xcomposite-egl, wayland-xcomposite-glx, webgl, xcb.

Aborted (core dumped)
```

@ziw-liu @edyoshikun I'm considering moving the waveorder imports into the function definitions as a deeper workaround. This would let us avoid the need to start and close napari for all but the `view` command. 

This is, admittedly, a poor workaround for the [core issue in `waveorder` #97](https://github.com/mehta-lab/waveorder/issues/97).

talonchandler:
Fixes #315.

With this PR I can run `recorder info` and `recorder convert` without a screen.

### DIFF
--- a/recOrder/scripts/cli.py
+++ b/recOrder/scripts/cli.py
@@ -2,14 +2,6 @@ import click
 import napari
 import numpy as np
 
-# Workaround waveorder #97
-# Create napari Viewer before other imports
-v = napari.Viewer()
-v.close()
-from recOrder.io.zarr_converter import ZarrConverter
-from waveorder.io import WaveorderReader
-
-
 @click.group()
 def cli():
     print(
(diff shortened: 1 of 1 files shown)


edyoshikun (approved):
Nicely done! I was able to run the converter and other functions that don't need the viewer in my local terminal and VScode SSH terminal. 
Thank you again!

PR #318, opened 2023-02-14
added option to parse through TS or DAC

### LINKED ISSUE #312: [BUG] LCs not recognized when using multiple Triggerscopes
For most setups, the use of one triggerscope so the current method for detecting and setting the voltages through the DAC by parsing `TS_DAC` from device properties works well. However, in the case of the Mantis where we have the triggerscope device named as `TS1_DAC` instead of `TS_DAC` our current string parsing fails. 

Check the following:
[main_widget.py](https://github.com/mehta-lab/recOrder/blob/0a175044050592a3f8af7a743234165e9f60bb38/recOrder/plugin/main_widget.py#L1331) and [Calibration.py](https://github.com/mehta-lab/recOrder/blob/0a175044050592a3f8af7a743234165e9f60bb38/recOrder/calib/Calibration.py#L184) will need to have some edits to recognize the triggerscope name.

edyoshikun:
There was no need to enumerate the devices as this reads the `State0` config properties and looks for devices with `TS`. As long as the channels have been set properly, the current implementation works well.

Additionally, there was no current need to add another dropbox for the DAC device, but perhaps in the future if users are are using DACs other than the triggerscope the if/else case should be changed to look for just `DAC` instead. 

Tested this on Mantis. Fixes #312

### DIFF
--- a/recOrder/plugin/main_widget.py
+++ b/recOrder/plugin/main_widget.py
@@ -1340,7 +1340,7 @@ class MainWidget(QWidget):
             memory = set()
             for i in range(cfg.size()):
                 prop = cfg.getSetting(i)
-                if "TS_DAC" in prop.getDeviceLabel():
+                if "TS" in prop.getDeviceLabel() and "DAC" in prop.getDeviceLabel():
                     dac = prop.getDeviceLabel()[-2:]
                     if dac not in memory:
                         self.ui.cb_lca.addItem("DAC" + dac)


ziw-liu (approved):
Code looks good. I just realized that the TriggerScope/DAC features are not documented at all. Maybe for another issue...

talonchandler (approved):

talonchandler:
Thanks @edyoshikun.  

> TriggerScope/DAC features are not documented at all

Thanks @ziw-liu. I'll move this to a separate issue.

edyoshikun:
I realized this will fail if the name of the Triggerscope is back to to 'TS' since the `calibration.py` in recOrder uses the name. Fixing this

ieivanov (changes requested):
This is not a permanent solution still. We should allow the users to name their DAQ device arbitrarily, and potentially use an [Arduino](https://micro-manager.org/Arduino) or [NI DAQ](https://micro-manager.org/NIDAQ) instead of a TriggerScope. DACs are registered as `SignalIODevice`. We should check for that using `mmc. getDeviceType`. Here is the SignalIODevice API: https://github.com/micro-manager/mmCoreAndDevices/blob/77ccd434cb84b86892b5544d1bf17aafa57cacdd/MMDevice/MMDevice.h#L881

talonchandler:
@edyoshikun will revisit this in preparation for polarization experiments on mantis.

PR #332, opened 2023-04-07
Multiprocessing script for reconstructing multiple positions and times

### LINKED ISSUE #299: [FEATURE] create example script which loops over positions and timepoints
It would be useful to have an example script which loops over positions and timepoints and reconstructs birefringence and phase. This script should set the standard of how multiple positions (e.g. `FOV1`, `FOV2`, `FOV3`, etc. or `A1`, `A2`, `A3`, etc.) from the same dataset (`2022_12_15 HEK`) are saved in a zarr store. This script can also show how to process multiple positions or timepoints in parallel using the `multiprocessing` package.

talonchandler:
Fixes #299.

This example script downloads the recOrder test dataset (if it hasn't already been downloaded), converts the MM dataset to zarr, then uses multiprocessing to read and write each position and time point into a new zarr store in parallel. 

I have confirmed a ~5.5-fold speedup processing this 6-volume (T=2, P=3) test dataset on the HPC login node. I was not able to get the same speedup on my local machine (weird macOS behavior with core handling?). 

Thanks @ziw-liu and @edyoshikun for your suggestions. I found that converting to zarr helped (as expected), and I found that using the HPC was essential...multiprocessing isn't working well locally.

@ziw-liu I'm happy to take suggestions for cleaning up this script. I'm still not sure how we can best deploy multiprocessing and slurm since in my experience they are so resource dependent and have so many moving parts.

@ieivanov this addresses the issue you opened. You're more than welcome to test and comment, but it's not strictly necessary since we can iterate on this script fairly quickly.

### DIFF
--- a/examples/parallel-reconstruct.py
+++ b/examples/parallel-reconstruct.py
@@ -66,6 +66,32 @@ def precomputation():
         width=X,
     )
 
+    # setup reconstructor.
+    reconstructor_args = {
+        "image_dim": (Y, X),
+        "mag": 20,  # magnification
+        "pixel_size_um": 6.5,  # pixel size in um
+        "n_slices": Z,  # number of slices in z-stack
+        "z_step_um": 2,  # z-step size in um
+        "wavelength_nm": 532,
+        "swing": 0.1,
+        "calibration_scheme": "4-State",  # "4-State" or "5-State"
+        "NA_obj": 0.4,  # numerical aperture of objective
+        "NA_illu": 0.2,  # numerical aperture of condenser
+        "n_obj_media": 1.0,  # refractive index of objective immersion media
+        "pad_z": 5,  # slices to pad for phase reconstruction boundary artifacts
+        "bg_correction": "global",  # BG correction method: "None", "local_fit", "global"
+        "mode": "3D",  # phase reconstruction mode, "2D" or "3D"
+        "use_gpu": False,
+        "gpu_id": 0,
+    }
+    reconstructor = initialize_reconstructor(
+        pipeline="QLIPP", **reconstructor_args
+    )
+
+    # reconstruct background Stokes
+    bg_stokes = reconstruct_qlipp_stokes(bg_data, reconstructor)
+
     # setup output zarr
     writer = open_ome_zarr(
         "./output/reconstructions_" + timestamp + ".zarr",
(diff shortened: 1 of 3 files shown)


mattersoflight:
>  I'm still not sure how we can best deploy multiprocessing and slurm since in my experience they are so resource dependent and have so many moving parts.

Let's consider this an internal deployment for now, i.e., meant for our internal collaborators. We can assume that these scripts will run on bruno with all its peculiarities.

examples/parallel-reconstruct.py
```diff
@@ -0,0 +1,171 @@
+import shutil
+import os
+import numpy as np
+import multiprocessing as mp
+from tqdm import tqdm
```
ziw-liu, line 5: Unused import

```suggestion

```

examples/parallel-reconstruct.py
```diff
@@ -0,0 +1,171 @@
...
+from wget import download
+from datetime import datetime
+from iohub import read_micromanager, open_ome_zarr
```
ziw-liu, line 8:
```suggestion
from iohub import open_ome_zarr
```

edyoshikun:
This code LGTM! I have a similar implementation that works on single node with 64 cores. I agree with @mattersoflight. I do think we should have a common place for our internal users to do the reconstructions via just multiprocessing in local machine or with slurm. I'm not sure where they should live, but for example I currently have [these](https://github.com/czbiohub/compmicro-sandbox/tree/master/zebrafishInfection/ed/slurm) in the sandbox. It's not the cleanest slurm implementation but it works for reconstructions.

examples/parallel-reconstruct.py
```diff
@@ -0,0 +1,170 @@
...
+    P = len(list(reader.positions()))
+    bg_data = load_bg(
+        os.path.join(data_folder, "2022_08_04_recOrder_pytest_20x_04NA/BG/"),
+        height=Y,
+        width=X,
+    )
+
+    # setup output zarr
+    timestamp = datetime.now().strftime("%Y%m%d_%H%M%S")
+    writer = open_ome_zarr(
```
ziw-liu, line 62: This script cannot be ran repeatedly without manual editing or removing previous data since the conversion cannot (and maybe should not) overwrite. So is it necessary to write to two different stores for conversion and reconstruction?

examples/parallel-reconstruct.py
```diff
@@ -0,0 +1,170 @@
...
+# result to a hcs-format zarr
+def single_process(i, reader, bg_data, writer, vol_start, vol_end):
+    T, C, Z, Y, X = reader["0/0/0"].data.shape
+    for vol in range(vol_start, vol_end + 1):
+        print("DEBUG:", i, vol)
+        p = int(np.floor(vol / T))
+        t = int(vol % T)
+        print(f"Reconstructing vol={vol}, pos={p}, time={t}")
+
+        data = reader["0/" + str(p) + "/0"]["0"][t, ...]
```
edyoshikun, line 93: I don't think we need the ellipsis here

examples/parallel-reconstruct.py
```diff
@@ -0,0 +1,170 @@
...
+
+        # Set up a reconstructor.
+        reconstructor_args = {
```
edyoshikun, line 96: I usually create the reconstructor outside of the process since all the reconstructions share the same parameters. I also do the bg_stokes reconstruction outside, because it's the same case.

examples/parallel-reconstruct.py
talonchandler: Good point. I just made a change to convert to a zarr store with a timestamped name instead of overwriting so that we can rerun the script. 

I know this isn't as space efficient as overwriting the converted store, but I think the clarity is worth it. 

I also added a comment mentioning that you should only use this download-and-convert section for a test run.

examples/parallel-reconstruct.py
talonchandler: Nope I 100% agree that that's the correct thing to do. Fixed!

ziw-liu (approved):

ieivanov:
Sorry for chiming in late on this.

From reading this script, I think it will work well and it's very powerful in that you can do parallel processing along time, position, or both, but I think it's overall a little complicated for an example script. Typically, large acquisitions would tend to be many time points over a few positions or many positions over a few time points such that doing parallel processing over one of these dimensions would use up your computational resources. We could provide an example script for both of these use cases, or maybe even just one as it would be easy to extend to the other. Again, this will not be as universal, but it will provide for a simple example.

Rather than `Process.join` I also prefer to use the `map` or `starmap` functions of `multiprocessing.Pool` along with `functools.partial` to quickly dispatch processors that apply a function over an iterable such as the T dimension of `data`.

I plan to implement multiprocessing of the mantis deskew in this way, if that approach seems nicer we can also transfer it here.

talonchandler:
> a little complicated for an example script

I agree! 

One of the challenges I've faced is having to choose/change between parallelizing over time, positions, or both. I'm betting this will be a increasingly common problem for us, so I'm all ears for simpler solutions. 

> I also prefer to use the `map` or `starmap` functions of `multiprocessing.Pool` along with `functools.partial`

Do you prefer these for syntax reasons (very valuable) or deeper reasons? 

Another (related?) challenge I've faced is the situation where you have more positions x times than cores, so you need to split them up. At one point I convinced myself with some off-the-cuff tests that throwing 1000 tasks at 10 cores via `Pool` was slower than divvying up the 1000 tasks into 100-task chunks that go to each core, but I'm not sure if I reached the right conclusion. 

If `map` (or alternatives) behave appropriately if you have more tasks than cores, then I can imagine a much simpler solution that switches between iterables: `range(T)`, `range(P)`, or `itertools.product(range(T), range(P))`. Definitely let me know what you come up with on mantis.

ieivanov:
>One of the challenges I've faced is having to choose/change between parallelizing over time, positions, or both. I'm betting this will be a increasingly common problem for us, so I'm all ears for simpler solutions.

I hear you. 

> Do you prefer these for syntax reasons (very valuable) or deeper reasons?

Largely for syntax reasons.

It's possible that there are intricacies with using `Pool`. It does have `imap`, `map_async`, and `imap_unordered` which are supposed to help with things like that. `imap` in particular can do the chunking you're describing for you. 

I didn't know about `itertools.product`. That along with `starmap` will make for pretty elegant-looking code, I think. Hopefully it will be performant too, let's test.

PR #335, opened 2023-04-10
Turn live mode off when loading calibration

### LINKED ISSUE #280: "Load calibration" when MM is live gives an error

talonchandler:
Fixes #280 

Tested on hummingbird

### DIFF
--- a/recOrder/plugin/main_widget.py
+++ b/recOrder/plugin/main_widget.py
@@ -1845,6 +1845,10 @@ class MainWidget(QWidget):
         def update_extinction(extinction):
             self.calib.extinction_ratio = float(extinction)
 
+        # Make sure Live Mode is off
+        if self.calib.snap_manager.getIsLiveModeOn():
+            self.calib.snap_manager.setLiveModeOn(False)
+
         # initialize worker properties for multi-threading
         self.ui.qbutton_stop_calib.clicked.connect(self.worker.quit)
         self.worker.yielded.connect(self.ui.le_extinction.setText)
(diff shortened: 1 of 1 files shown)


ziw-liu:
To my eyes this is very similar to #282 but glad this worked! @talonchandler maybe also consider using that context manager?

talonchandler:
It looks like we're using the context manager when we capture a background, and we're using these lines for running a calibration (where I got this from) and now for loading a calibration:
```
if self.calib.snap_manager.getIsLiveModeOn():
    self.calib.snap_manager.setLiveModeOn(False)
```

I agree that the context manager should do the same thing, so I don't understand why #282 isn't working. 

Do you think it's worth the effort to test and understand this? Or do you think we should merge this fix and handle later?

ziw-liu:
It seems like #282 didn't work because [the `load_calibration` worker](https://github.com/mehta-lab/recOrder/blob/fbbd7c8b44ea0bdd7b7cb159c390277c2983fd1e/recOrder/calib/calibration_workers.py#L439) makes MMCore calls directly instead of reusing existing methods of the `Calibration` object. I can fix that.

ziw-liu (approved):
@talonchandler I'm fine with this PR either way since its just a few lines. It's your call whether we need this quick patch for the 0.3 release, but if you decide to merge maybe comment with `FIXME` in the code so that we don't forget about it.

talonchandler:
Thanks for your flexibility @ziw-liu. I've added FIXME comments, and I'm looking forward to moving the MM calls off of the main thread later.

PR #336, opened 2023-04-11
Add overlay option to `view` command

talonchandler:
`iohub` is not planning direct support for saving and displaying RGB images (https://github.com/czbiohub/iohub/issues/74), so this PR provides a recOrder-specific way to create quick overlays. 

This adds an `-o` option to `recorder view` that adds an overlay layer, identical to what you see after an acquisition from the `recOrder` plugin. 

In the short-medium term I might be the only user for this option, but it's definitely a common use case for me.

### DIFF
--- a/recOrder/io/utils.py
+++ b/recOrder/io/utils.py
@@ -282,91 +282,71 @@ def generic_hsv_overlay(
     return overlay_final[0] if mode == "2D" else overlay_final
 
 
-def ret_ori_overlay(
-    retardance, orientation, ret_max=10, mode="2D", cmap="JCh"
-):
+def ret_ori_overlay(retardance, orientation, ret_max=10, cmap="JCh"):
     """
     This function will create an overlay of retardance and orientation with two different colormap options.
     HSV is the standard Hue, Saturation, Value colormap while JCh is a similar colormap but is perceptually uniform.
 
     Parameters
     ----------
-    retardance:             (nd-array) retardance array of shape (N, Y, X) or (Y, X) in nanometers
-    orientation:            (nd-array) orientation array of shape (N, Y, X) or (Y, X) in radian [0, pi]
-    ret_max:                (float) maximum displayed retardance.  Typically use adjusted contrast limits.
-    mode:                   (str) '2D' or '3D'
+    retardance:             (nd-array) retardance array in nanometers (shape must match orientation)
+    orientation:            (nd-array) orientation array in radian [0, pi] (shape must match retardance)
+    ret_max:                (float) maximum displayed retardance. Typically use adjusted contrast limits.
     cmap:                   (str) 'JCh' or 'HSV'
 
     Returns
     -------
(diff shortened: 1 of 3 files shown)


recOrder/scripts/cli.py
```diff
@@ -1,6 +1,9 @@
 import click
 import napari
+
+v = napari.Viewer()  # open viewer right away to use on hpc
```
talonchandler, line 4: this earlier issue https://github.com/mehta-lab/recOrder/issues/315 popped up again, so I needed to create the viewer at the top

recOrder/io/utils.py
```diff
@@ -282,91 +282,71 @@ def generic_hsv_overlay(
     return overlay_final[0] if mode == "2D" else overlay_final
 
 
-def ret_ori_overlay(
-    retardance, orientation, ret_max=10, mode="2D", cmap="JCh"
-):
+def ret_ori_overlay(retardance, orientation, ret_max=10, cmap="JCh"):
```
ziw-liu, line 285: This function is used in acquisition too:
https://github.com/mehta-lab/recOrder/blob/107e4c6685ed15fdab7dc4c6ffc80b5cfe901a2a/recOrder/plugin/main_widget.py#L1155-L1173

Is the plan here to stop distinguishing between 2D and (the much slower) 3D overlays?

recOrder/scripts/cli.py
```diff
@@ -174,7 +197,18 @@ def cli():
...
+    "-o",
+    is_flag=True,
+    default=False,
+    help="""
+    Computes a retardance-orientation HSV overlay layers. Requires channel 
+    names `Retardance` and `Orientation` to be present, and requires `--layers 
+    position`
+    """,
+)
+def view(filename, positions=None, layers=None, overlay=False):
```
ziw-liu, line 211: My impression is that default value of arguments in the signature are not need in click options if the decorators already have them (and they are different here).

talonchandler: Thanks!

recOrder/io/utils.py
ziw-liu: Then maybe remove the outdated argument from the main widget method?

talonchandler: Fixed!

recOrder/plugin/main_widget.py
```diff
@@ -1160,7 +1159,6 @@ def handle_bire_image_update(self, value: NDArray):
             retardance=value[0],
             orientation=value[1],
             ret_max=np.percentile(value[0], 99.99),
-            mode=self.acq_mode,
```
ziw-liu, line 1163: How about the `show_info` statement below? If the new 3D overlay implementation is still significantly slower than 2D I guess we still want it around?

talonchandler: Thanks for mentioning it. During my testing today the overlay was noticeably faster than before but still had a noticeable delay, so I still like having the `show_info` statement.

ziw-liu (approved):
Looks good!

PR #337, opened 2023-04-11
Slurm-scripts

edyoshikun:
This PR adds examples on how to use slurm for parallel reconstructions for birefringence and phase. 

Both pipilines convert the data to zarr, create an empty zarr store for the output and parallelize the reconstructions per position to simultaneously write to the output zarr.

These scripts run using our zenodo dataset, but bigger datasets really take an advantage using HPC.

### DIFF
--- a/examples/slurm-scripts/bire_recon/empty_zarr.py
+++ b/examples/slurm-scripts/bire_recon/empty_zarr.py
@@ -1,0 +1,69 @@
+# %%%
+from iohub.reader import read_micromanager
+from iohub.ngff import open_ome_zarr
+
+# debugging
+from tqdm import tqdm
+import click
+import numpy as np
+from datetime import datetime
+
+
+@click.group()
+def cli():
+    pass
+
+
+@cli.command()
+@click.help_option("-h", "--help")
+@click.option(
+    "--input",
+    required=True,
+    type=click.Path(exists=True),
+    help="path to the RAW Zarrstore",
+)
+@click.option(
+    "--output",
+    required=True,
+    type=str,
+    help="path to save the bire + fluor dataset (i.e /tmp)",
+)
+def make_empty_array(input, output):
+    reader = read_micromanager(input)
+    T, C, Z, Y, X = reader.shape
+    num_positions = reader.get_num_positions()
+    channel_names = reader.channel_names
+    non_state_indices = [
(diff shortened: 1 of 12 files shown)


edyoshikun:
I think I need to add documentation on how to run the code. 
The scripts are bash scripts and the phase or birefringence reconstructions can be executed by `bash batch_bire_recon.sh` or `bash batch_phase_recon.sh` respectively. Thank you both!

mattersoflight:
Thanks for these utilities @edyoshikun.  Yes, a quick how-to on staging the data will be useful. These can be comments in specific scripts or a markdown guide linked in the [docs](https://github.com/mehta-lab/recOrder/blob/main/docs/README.md).

Have you already implemented a utility that allows the reconstruction of data parallelized by positions, preceded by or followed by, the creation of metadata at all positions within the zarr store? Is `iohub`'s metadata API now modular enough to do so?

ziw-liu:
> Have you already implemented a utility that allows the reconstruction of data parallelized by positions, preceded by or followed by, the creation of metadata at all positions within the zarr store? Is `iohub`'s metadata API now modular enough to do so?

I think [this file](https://github.com/mehta-lab/recOrder/blob/7090a795664b317e3fca8f876d91a783ec6bbec8/examples/slurm-scripts/bire_recon/empty_zarr.py) does this (although using iohub internals which are now wrapped in public API, see following comment).

examples/slurm-scripts/bire_recon/empty_zarr.py
```diff
@@ -0,0 +1,71 @@
...
+        layout="hcs",
+        mode="w-",
+        channel_names=recon_chan_names,
+    ) as dataset:
+        # Make the positions
+        for i in tqdm(range(num_positions)):
+            pos = dataset.create_position("0", "0", str(i))
+            # this is a 'hack' to create placeholder metadata
+            # future iohub should expose a public API for emtpy images
+            pos._create_image_meta("0")
```
ziw-liu, line 60: Consider using `pos.create_zeros(...)` instead

edyoshikun: thanks! I will use this from now on.

talonchandler:
@edyoshikun this PR no longer needs a review, correct? Can you change to draft or close?

PR #344, opened 2023-04-18
Minor documentation updates for 0.3.0

### LINKED ISSUE #314: Remove separate arm64 installation documentation
In `recOrder` 0.2.x a x86_64 Python environment is required for napari/PyQt5 to work on macOS. With #313 we can remove those instructions in the next release since the same installation process now works for both x86_64 and arm64 machines.

talonchandler:
Fixes #314 and other minor typos.


ziw-liu (approved):
Thanks for fixing these!

talonchandler:
> Please point to [microscope installation guide](https://github.com/mehta-lab/recOrder/blob/main/docs/microscope-installation-guide.md) from README.

Good call, fixed. 

> @ieivanov I see that Mark Tsuchida gave a go-ahead to merge the MeadolarkLC device adapter with nightly. Is it ready to merge from your point of view? Should recOrder 0.3.0 depend on it?

Following our discussions, I think we should release `0.3.0` and we can depend on the new device adapter for `0.4.0`.

PR #345, opened 2023-04-19
Pin README links

talonchandler:
Similar to `https://github.com/mehta-lab/waveorder/pull/116`, two of the links on PyPI are broken. This PR pins the links. I think this issue is minor enough that is doesn't justify a rerelease, but I'm fixing this now so that we don't forget for the next release.


ziw-liu (approved):
Thanks!

PR #346, opened 2023-04-21
Test docking the plugin widget

ziw-liu:
Checks that the GUI can be initialized and docked to the napari viewer without throwing an error.


ziw-liu:
Already failing on GitHub CI!

ziw-liu:
Couldn't get Qt to work on the Linux runners. But Windows and macOS runs fine. I think as long as it tests on Windows (where the microscopes are) it's of some value.

talonchandler (approved):
Looks great! Thanks for doing this @ziw-liu.

PR #353, opened 2023-04-25
Connect background corrections

### LINKED ISSUE #349: Fix background correction path problems
In my initial prototypes my background corrections strings weren't being passed correctly. After discussing with @ziw-liu and @edyoshikun I think I need to use `str` type and validate. I expect this to be a fairly shallow bug.

talonchandler:
Fixes #349. 

@edyoshikun I couldn't get `DirectoryPath` to work in the way you suggested, but I ended up finding a fairly simple method with `os.path.isdir`. 

I've tested on hummingbird.


edyoshikun (approved):
this LGTM!

PR #355, opened 2023-04-26
Converter cleanup

ziw-liu:
Remove remaining converter util function.


ziw-liu:
@talonchandler there seems to be a syntax error in the CLI test that is causing CI to fail here. But I'll leave that to you.

talonchandler:
Thanks!

PR #357, opened 2023-04-27
Draw birefringence overlay on napari layers event

### LINKED ISSUE #348: No simple and uniform way to visualize orientation and retardance-orientation overlays
@ziw-liu @mattersoflight and I discussed a few possible direction for recOrder visualizations: 

**Current behavior:** reconstructions can be dragged and dropped into napari (thanks `iohub`!) , but there isn't an easy way to color the orientation channel or generate overlays. 

**Possible solutions:**
1) **Use a custom reader plugin** - this won't allow us to generate overlays, so it won't work by itself. 
2) **Write RGB overlays alongside the reconstructions *and* use a custom reader plugin** - this meets our requirements, but requires a lot of unnecessary data to be stored. 
3) **Use a CLI like `recOrder view` to generate RGB overlays in a napari window** - this will work, and it could be unified with the `recOrder` GUI, but it requires a user to use the CLI...no drag and drop option. 
4) @ziw-liu's latest idea - **Use the napari event loop to generate an overlay whenever layers named "Retardance" and "Orientation" are added to the viewer with the `recOrder-napari` plugin running**. This solution allows drag-and-drop visualization, and it seems intuitive to me that this will only work when you have the `recOrder` plugin open. This approach unifies our visualizations, doesn't bump our storage requirements, works for arbitrarily shaped arrays, and is a reasonably non-invasive change to make. 

I really like option 4. A potential drawback is that it seems a bit "magic" (where did that overlay come from and why is it happening now?), but we might alleviate this by printing messages to the user.

ziw-liu:
Fix #348.

Currently this does not delay the compute and brings all the data into memory. So it will NOT work with large datasets.

Making it work lazily through Dask needs some changes to the overlay rendering function, which uses `np.copy()` that is not supported by Dask arrays.

### DIFF
--- a/recOrder/io/utils.py
+++ b/recOrder/io/utils.py
@@ -1,3 +1,4 @@
+from typing import Literal
 import glob
 import logging
 import os
@@ -254,7 +255,9 @@ def generic_hsv_overlay(
     return overlay_final[0] if mode == "2D" else overlay_final
 
 
-def ret_ori_overlay(retardance, orientation, ret_max=10, cmap="JCh"):
+def ret_ori_overlay(
+    retardance, orientation, ret_max=10, cmap: Literal["JCh", "HSV"] = "JCh"
+):
     """
     This function will create an overlay of retardance and orientation with two different colormap options.
     HSV is the standard Hue, Saturation, Value colormap while JCh is a similar colormap but is perceptually uniform.
@@ -273,7 +276,8 @@ def ret_ori_overlay(retardance, orientation, ret_max=10, cmap="JCh"):
     """
     if retardance.shape != orientation.shape:
         raise ValueError(
-            f"Retardance and Orientation shapes do not match: {retardance.shape} vs. {orientation.shape}"
+            "Retardance and Orientation shapes do not match: "
+            f"{retardance.shape} vs. {orientation.shape}"
         )
 
     # Prepare input and output arrays
--- a/recOrder/io/utils.py
+++ b/recOrder/io/utils.py
@@ -282,9 +282,8 @@ def ret_ori_overlay(
 
     # Prepare input and output arrays
     ret_ = np.clip(retardance, 0, ret_max)  # clip and copy
-    ori_ = (
-        np.copy(orientation) * 360 / np.pi
-    )  # convert 180 degree range into 360 to match periodicity of hue.
+    # Convert 180 degree range into 360 to match periodicity of hue.
+    ori_ = orientation * 360 / np.pi
     overlay_final = np.zeros_like(retardance)
 
     # FIX ME: this binning code leads to artifacts.
--- a/recOrder/plugin/main_widget.py
+++ b/recOrder/plugin/main_widget.py
@@ -1168,6 +1168,10 @@ class MainWidget(QWidget):
             )
             self._draw_bire_overlay()
         if event.source[-1].name == channels[1]:
+            logging.info(
+                "Detected orientation layer in updated layer list."
+                "Setting its colormap to HSV"
+            )
             self.viewer.layers[channels[1]].colormap = "hsv"
 
     def _draw_bire_overlay(self):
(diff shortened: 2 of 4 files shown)


ziw-liu:
Tests won't pass because of wrong waveorder dependency in `setup.cfg`. Should we update that now or just ignore it?

ziw-liu:
@talonchandler Is explicit `np.copy()` necessary here?

https://github.com/mehta-lab/recOrder/blob/88d7c98515a037052001f624847fc3f7e7e21f2b/recOrder/io/utils.py#L286

talonchandler:
Just tested, and this is looking really great. Thanks for the fast turnaround @ziw-liu. 

Working:
- drag-and-dropping the `reconstruction.zarr` files generates overlays automatically when `recOrder-napari` is open. Fantastic! And very intuitive. 

Needs discussion:
- This line makes `recOrder` acquisitions open the layers with a `2D` or `3D` suffix, so the overlays don't show up during living acquisitions. We could either (1) remove the suffix, or (2) match for ["Retardance*", "Orientation*"] layers names so that the overlays are generated for online acquisitions. I lean towards (1) (i.e. remove ` + self.acq_mode` from this line)  
https://github.com/mehta-lab/recOrder/blob/30112585e6c5ab2d69459fdfea2a8abae9ff25c2/recOrder/plugin/main_widget.py#L1170

- Can I request that we arrange the logic so that the newly generated overlays always appear as the top layer in napari? I briefly tested the acquisition after I removed the layer-name suffix, and a new pair of "Retardance" and "Overlay" layers triggers in-place repopulation of the existing layer---can the layer be deleted then regenerated instead?

- Can I request that any newly added "Orientation" layer be shown in "HSV"?

ziw-liu:
> * This line makes `recOrder` acquisitions open the layers with a `2D` or `3D` suffix, so the overlays don't show up during living acquisitions. We could either (1) remove the suffix, or (2) match for ["Retardance*", "Orientation*"] layers names so that the overlays are generated for online acquisitions. I lean towards (1) (i.e. remove ` + self.acq_mode` from this line)

I agree that we should remove the suffix. Not just because it is easier for this PR, but also for consistency with how the data is actually stored.

ziw-liu:
> * Can I request that any newly added "Orientation" layer be shown in "HSV"?

ec49eb85760d6c427cb5993054c810c7cd6e8013

talonchandler:
> Can I request that any newly added "Orientation" layer be shown in "HSV"? 

Thanks! The HSV colormap for new layers named "Orientation" is working well. 

This demo shows the overlay's behavior after repeated acquisition...the first acquisition puts the overlay on top, then subsequent acquisitions puts it at the bottom. 
recOrder-demo-overlays

Another design decision point/questions has come up during testing. If I drag two `reconstruction.zarr` folders into napari, the first one will generate overlays as expected, but the second one will not generate overlays because napari adds the layers as "Retardance [1]" and "Orientation [1]". Dragging the same zarr in two times creates different behaviors which is a bit confusing. 

Might it be possible to trigger the overlay based on the "raw layer name" before napari appends the "[1]"? Or if that's impossible, maybe we're forced into implementing some sort of glob-matching logic?

ziw-liu:
> If I drag two `reconstruction.zarr` folders into napari, the first one will generate overlays as expected, but the second one will not generate overlays because napari adds the layers as "Retardance [1]" and "Orientation [1]". Dragging the same zarr in two times creates different behaviors which is a bit confusing.
> 
> Might it be possible to trigger the overlay based on the "raw layer name" before napari appends the "[1]"? Or if that's impossible, maybe we're forced into implementing some sort of glob-matching logic?

This can become complicated:

1. We don't have the layer data before napari adds the layers. There's no image to compute the overlay.
2. The multiple overlay layers need to get named somehow to distinguish them.

I think loading multiple FOVs as repetitive sets of layers should generally be avoided, given how the layer control becomes less ergonomic and how the performance suffers from too many layers.

PR #359, opened 2023-04-28
Multiple birefringence overlays

ziw-liu:
Implements https://github.com/mehta-lab/recOrder/pull/357#issuecomment-1526522379.

This works on a best-effort basis. If the user deletes part of the layer list (e.g. keeping orientation but not retardance) and loads another dataset, this will produce wrong overlays.

Also this treats any layer whose name ***starts with*** 'Retardance' and 'Orientation' as a birefringence layer. So layers with names like 'Retardance-mask' and 'Orientation of the fish body axis' will also be treated as birefringence layers and cause errors.

### DIFF
--- a/recOrder/plugin/main_widget.py
+++ b/recOrder/plugin/main_widget.py
@@ -1181,7 +1181,9 @@ class MainWidget(QWidget):
                         "Detected updated birefringence layers: "
                         f"'{latest_layer_name}', '{other_name}'"
                     )
-                    self._draw_bire_overlay(overlay_name)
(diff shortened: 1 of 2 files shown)


ziw-liu:
recorder-multi-overlay

talonchandler (approved):
This is fantastic! Thanks again @ziw-liu, and I apologize again that I've been slow to get to this. 

This functionality makes dragging-and-dropping multiple zarr stores into napari for comparison extremely easy and intuitive. 

Tested on hummingbird.

Merging.

PR #360, opened 2023-05-02
Multiprocessing 0.4.0dev with slurm/bash

edyoshikun:
This PR adds multiprocessing for reconstruction of birefringence only, bire and phase, and phase only for the 0.4.0 dev version. This PR solves #352

### DIFF
--- a/recOrder/cli/apply_inverse_transfer_function.py
+++ b/recOrder/cli/apply_inverse_transfer_function.py
@@ -4,11 +4,6 @@ import torch
 import yaml
 from iohub import open_ome_zarr
 from recOrder.cli.printing import echo_headline, echo_settings
-import torch.multiprocessing as mp
-import glob
-from tqdm import tqdm
-import os
-
 from recOrder.cli.settings import (
     TransferFunctionSettings,
     ApplyInverseSettings,
@@ -17,7 +12,6 @@ from recOrder.cli.parsing import (
     input_data_path_argument,
     config_path_option,
     output_dataset_options,
-    cores_option,
 )
 from recOrder.io import utils
 from waveorder.models import (
@@ -26,7 +20,38 @@ from waveorder.models import (
     phase_thick_3d,
 )
 
-MULTIPROCESSING_FLAG = True
+import glob
+import os
+import multiprocessing as mp
+from tqdm import tqdm
+
+# import torch.multiprocessing as mp
+
+
+def parallel(func=None, **options):
+    if func is None:
+        return functools.partial(parallel, **options)
+
+    def wrapper(*arguments):
+        print(f'arguments {arguments}')
+        pool = mp.Pool(mp.cpu_count())
+        #Split the arguments
+        position_paths = arguments[0]
+        recon_params = arguments[1:]
+
+        # Multiprocessing
+        result = []
+        for position in position_paths:
+            print(f"filepath: {position}")
+            args = (position,) + recon_params
+            print(f'args: {args}')
+            for results in tqdm(pool.starmap(func, [args])):
+                result.append(results)
+        pool.close()
+        pool.join()
+        return result
+
+    return wrapper
 
 
 def _check_shape_consistency(settings, data_shape):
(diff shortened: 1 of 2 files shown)


talonchandler:
Fantastic! I'm really excited to try this @edyoshikun. 

One small heads up...can you change this PR to merge to `0.4.0dev` instead of `main`? That should make it easier to see the changes you've made.

edyoshikun:
@talonchandler  good catch! Changed to merge to 0.4.0dev. 

I am still working on adding the `-j` flag for multiprocessing and thinking about the logic for the wildcards. The current implementation if provided a path to a position i.e `recorder apply-inverse-transfer-function /path/to/.zarr/0/0/0  /path/to/transfer_function.zarr /output/path/zarr` works nice. 

However, for the wildcard situation, we would need to change the logic for how we create the empty array such that we can have the same file tree structure as the input.  I assume that is what we want. At the moment, it can write multiple positions in one zarr store but it does not mirror the original structure.

edyoshikun:
After our discussion @talonchandler, I made the code such that when the user uses the wildcard the input file tree is mirrored that way we preserve the Plate, Well, and FOV information. i.e `recorder apply-inverse-transfer-function ./store_path.zarr/*/*/*  ./TF.zarr -o ./output_store.zarr` will create the same structure as the input. If we select only a limited number of positions, the store will only create those empty positions that way we are not writing nonsense empty arrays. 

I think this is ready to review now.

talonchandler (commented):
Thanks for your patience on this review, @edyoshikun. 

I think there are (at least!) two fantastic pieces here that are ripe for reuse:
- multiprocessing over positions in a clear way
- the file handling with `nargs=-1` to let the user select subsets of the positions (this is an excellent implementation, thanks!)

My main concern is about the duplication of logic in the `apply_inverse_transfer_function_cli` function and the `multiprocess_position` function. This mirroring is best illustrated by the `mp_flag` parameter which is an integer from 0-5 that acts as a switch between the reconstruction paths. This flag is duplicating logic, and I think this duplication is pointing us towards simplification. Can we find a solution that doesn't use switches? And allows us to reuse your code elsewhere?

----
The code in this PR currently has the following structure:
```
def multiprocess_position(...):
    the function that is processed on a single core

def apply_inverse_transfer_function_cli(...):
    1. file list handling (for nargs=-1 stuff)

    2. the logic that builds the multiprocessing queue
    (before this PR, this function contained all of the switching/dispatching logic 
    to match recOrder to waveorder, and now some of this logic is duplicated inside this 
    function and the multiprocess_position function)
    
    3. starting the multiprocessing Pool
    
def apply_inverse_transfer_function(...):
    a extremely thin wrapper for the cli
```
---
Would the following redesign work? Is this in improvement in modularity/maintainability?

```
def apply_inverse_transfer_function_cli(...):
   leave this function unchanged...this is where the densest recOrder-waveorder logic lives, 
    and I'd like to keep it in a single function. This function might need to be renamed.
    
def apply_inverse_transfer_function(...):
    1. file list handling (for nargs=-1 stuff) --- this can be copied from what you already have and/or moved to it's own utils function for reuse
    2. building the multiprocessing queue of `apply_inverse_transfer_function_cli` calls --- this shouldn't need to duplicate any logic, right?
    3. starting the multiprocessing Pool    
```

My ultimate goal here is to abstract out the work in this PR so that it can be applied easily to other CLI tools (especially mantis). I think we can get to a point where we can easily add multiprocessing over positions to any CLI function with a decorator.

talonchandler:
I just looked into multiprocessing with decorators. I'm seeing in a few places that there are pickling-related reasons why it won't work with the syntax sugar `@` symbol, but it will work with function wrapping. 

This is an example of what I have in mind:
https://stackoverflow.com/a/67137738

edyoshikun:
Sorry for the delay @talonchandler. This is a refactor to what I should've done probably from the beginning. Thank you for the input. 

I implemented something similar to the stack overflow you sent. I added a decorator for Pool, so that each position applies the `apply_inverse_transfer_function_cli` in its own process. Here I had to modify the click `apply_inverse_transfer_function` so that we do the following:

- Do the logic to parse the single position or wildcard .
- Create an empty array that mirrors the input path (single position | wildcard)
- Parallelize per position using the decorator.

I think this is a lot cleaner and simplified.

talonchandler:
I haven't forgotten about this and #337! Let's revisit these once we find an `iohub`-supported way to process multiple positions with slurm.

talonchandler:
Following @edyoshikun's [comment here](https://github.com/czbiohub-sf/mantis/pull/46#issuecomment-1588280845), we may revisit this after making changes to support slurm. Changing to draft for now.

edyoshikun:
Closing this PR in place of #380

PR #377, opened 2023-06-13
Refactor and test compute transfer function

clinton-huynh:
Added test cases for compute_transfer_function.
Organized logic in compute_transfer function.
Added documentation for compute_transfer_function.


recOrder/tests/cli_tests/test_cli.py
```diff
@@ -1,10 +1,37 @@
...
+    for option in ("-c", "--config-path"):
+        cmd = "compute-transfer-function" + option
+        result = runner.invoke(cli, cmd)
+        assert result.exit_code == 2
+        assert "Error" in result.output
+
+def test_compute_transfer_output_file():
+    runner = CliRunner()
+    path = "test"
+    # for option in ("-o", "--output-path"):
```
talonchandler, line 32: It would be nice to get this to work. Feel free to post your error msg and we can help out.

recOrder/cli/compute_transfer_function.py
```diff
@@ -34,76 +45,10 @@ def compute_transfer_function_cli(config_path, output_path):
...
-        intensity_to_stokes_matrix = (
-            inplane_anisotropic_thin_pol3d.calculate_transfer_function(
-                **settings.birefringence_transfer_function_settings.dict()
-            )
-        )
-        # Save
-        dataset[
-            "intensity_to_stokes_matrix"
-        ] = intensity_to_stokes_matrix.cpu().numpy()[None, None, None, ...]
+        dataset = generate_birefringence(settings, dataset)
```
talonchandler, line 48: I think this refactor is great!

As we discussed, I don't think you need to return `dataset` here. I also think you can rename `generate_birefringence` to `generate_birefringence_transfer_function`.

recOrder/cli/compute_transfer_function.py
```diff
@@ -34,76 +45,10 @@ def compute_transfer_function_cli(config_path, output_path):
...
-            # Save
-            dataset[
-                "real_potential_transfer_function"
-            ] = real_potential_transfer_function.cpu().numpy()[None, None, ...]
-            dataset[
-                "imaginary_potential_transfer_function"
-            ] = imaginary_potential_transfer_function.cpu().numpy()[
-                None, None, ...
-            ]
+        datatset = generate_phase(settings, dataset)
```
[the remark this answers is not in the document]
talonchandler: See me comment above...the typo isn't actually a bug because you don't need to return dataset. 

I'll also suggest `generate_phase_transfer_function`.

test/.zattrs
```diff
@@ -0,0 +1,117 @@
+{
+    "multiscales": [
```
talonchandler, line 2: I think these generated files can be removed from the git repo. Feel free to add to `.gitignore` if it's more convenient.

talonchandler (commented):
Excellent work @clinton-huynh! I really like the functions you refactored, and the tests are extremely useful since this is a critical piece of code for `recOrder`. 

Following our discussion, I left specific suggestions in the comment below. I will also suggest a few more tests:
- test with user-edited config settings, particularly the universal settings
- test that the correct transfer functions names are written to file after using `reconstruct_phase = True` and `reconstruct_birefringence = True`.

recOrder/tests/cli_tests/test_cli.py
clinton-huynh: I realized that spacing does matter when invoking the calls. Proper spacing fixed the --output-path not working.

recOrder/cli/compute_transfer_function.py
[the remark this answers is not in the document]
clinton-huynh: I removed the return, and renamed the functions to generate_and_save_phase_transfer_function and generate_and_save_birefringence_transfer_function.

test/.zattrs
clinton-huynh: I'm not quite sure what proper steps to take to remove the files. Could we discuss this and .gitignore further?

talonchandler: From the command line, you can stop tracking a file with `git rm file.txt` then `git commit`. It's typically `right-click -> delete` in git GUIs. 

You can also add a specific file name or glob to the `.gitignore` file to ignore a file.

It's a bit tricky. We can chat later this morning if you can't get it to work.

recOrder/cli/compute_transfer_function.py
```diff
@@ -30,82 +41,19 @@ def compute_transfer_function_cli(config_path, output_path):
...
-            # Save
-            dataset[
-                "real_potential_transfer_function"
-            ] = real_potential_transfer_function.cpu().numpy()[None, None, ...]
-            dataset[
-                "imaginary_potential_transfer_function"
-            ] = imaginary_potential_transfer_function.cpu().numpy()[
-                None, None, ...
-            ]
+        generate_save_phase_transfer_function(settings, dataset)
```
talonchandler, line 53: Can you make this consistent with `generate_and_save_birefringence_transfer_function`? So ``generate_and_save_phase_transfer_function`.

recOrder/tests/cli_tests/test_cli.py
```diff
@@ -1,10 +1,109 @@
...
+    mock_settings.universal_settings.dict.return_value = [{
+        'reconstruct_birefringence': False,
+        'reconstruct_phase': True,
+        'reconstruct_dimensions': 2,
+        'wavelength_illumination': 0.532
+    }
+    ]
+    mock_settings.universal_settings.reconstruct_birefringence = False
+    mock_settings.universal_settings.reconstruct_phase = True
+    mock_settings.universal_settings.econstruct_dimension = 2
```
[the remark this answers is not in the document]
talonchandler: New issue here: https://github.com/mehta-lab/recOrder/issues/379

recOrder/cli/compute_transfer_function.py
```diff
@@ -13,8 +13,19 @@
 
 
 def compute_transfer_function_cli(config_path, output_path):
+    """CLI command to compute the transfer function given a configuration file path 
+    and a desired output path. Given no arguments, a default configuration file will 
+    be used and the output will be in transfer-function.zarr.
+
+    Parameters
+    ----------
+        config_path : string
```
ziw-liu, line 22: Extra indentation here ([style guide](https://numpydoc.readthedocs.io/en/latest/format.html)).

talonchandler (approved):
Looks great to me! Thanks a bunch @clinton-huynh. 

@ziw-liu I'll let you merge in case you want to take one more look.

recOrder/cli/compute_transfer_function.py
```diff
@@ -13,8 +13,19 @@
...
+    Parameters
+    ----------
+    config_path : string
+        Path of the configuration file.
+    output_path : string
+        Path of the output file.
+    """
     # Load config file
-    if config_path is None:
+    if config_path is None or "None":
```
ziw-liu, line 28: When will this be a literal `"None"`?

clinton-huynh: I believe Talon wanted the function to take in a literal "None" just in case.

recOrder/cli/compute_transfer_function.py
```diff
@@ -115,6 +60,96 @@ def compute_transfer_function_cli(config_path, output_path):
...
+    dataset[
+        "intensity_to_stokes_matrix"
+    ] = intensity_to_stokes_matrix.cpu().numpy()[None, None, None, ...]
+
+def generate_and_save_phase_transfer_function(settings, dataset):
+    """Generates and saves the phase transfer function to the dataset, based on the settings.
+
+    Parameters
+    ----------
+    settings: object
```
ziw-liu, line 95: Annotating non-builtin types with `object` is an unfortunate legacy this code base inherited from Java developers. In Python we should use the actual type in both type hints and docstrings. In this case `TransferFunctionSettings`.

ziw-liu: See an example here:
https://github.com/mehta-lab/recOrder/blob/6f8f05f500c9298053caeeba6f35788b738317e4/recOrder/acq/acq_functions.py#L120-L143

clinton-huynh: Sounds good. I'll make those changes. I noticed that they put the default arguments into the parameters while I have mine in the description/summary of compute_transfer_function_cli. Should I change that to match as well?

recOrder/cli/compute_transfer_function.py
ziw-liu: I don't think this case should be allowed. If a user wants to trigger the default settings explicitly we can add a `"default"` option.

ziw-liu: I don't think this case should be allowed. If a user wants to trigger the default settings explicitly we can add a `"default"` option.

recOrder/cli/compute_transfer_function.py
ziw-liu: In `compute_transfer_function_cli` you can't really move the default into the signature. The argument takes a str path, but the defualt (or actually fallback) behavior does not read the settings from any file. This is ok as long as it is documented.

recOrder/cli/compute_transfer_function.py
talonchandler: Sorry I made a misleading suggestion. I agree that we should drop the `or "None":` here. 

My (very minor) concern that I mentioned in person to Clinton was that the "recreation strings" can be incorrect if you don't provide an argument. 

So `recorder compute-transfer-function` would print:
```
Recreate this transfer function with:
>> recorder compute-transfer-function None -o ./transfer-function.zarr
```
which won't actually run. 

I can quickly fix this.

recOrder/tests/conftest.py
```diff
@@ -0,0 +1,38 @@
...
+from iohub.ngff import open_ome_zarr
+
+@pytest.fixture(scope="function")
+def setup_default_ctf_settings(tmp_path):
+    settings = TransferFunctionSettings()
+    dataset = open_ome_zarr(tmp_path, layout="fov", mode="w", channel_names=["None"])
+    yield settings, dataset
+
+@pytest.fixture(scope="function")
+def setup_b_false_ctf_settings(tmp_path):
```
ziw-liu, line 12: Please use a more descriptive name for birefringence and phase instead of 'b' and 'p'.

ziw-liu: Also 'ctf' is not something obvious. If you think the function name is too long you can remove 'setup_'.

Since you are yielding a tuple of (settings, OTF) probably reorder it to reflect the sequence.

recOrder/tests/cli_tests/test_cli.py
```diff
@@ -1,10 +1,154 @@
 from recOrder.cli.main import cli
+import unittest
```
ziw-liu, line 2: Unused import?

recOrder/tests/cli_tests/test_cli.py
```diff
@@ -1,10 +1,154 @@
...
+    assert result.exit_code == 0
+    assert "Generating" in result.output
+
+def test_compute_transfer_blank_config():
+    runner = CliRunner()
+    for option in ("-c ", "--config-path "):
+        cmd = "compute-transfer-function " + option
+        result = runner.invoke(cli, cmd)
+        assert result.exit_code == 2
+        assert "Error" in result.output
```
ziw-liu, line 40: Here you are testing that if no config is provided it will raise an error. Yet in this line:
https://github.com/mehta-lab/recOrder/blob/6f8f05f500c9298053caeeba6f35788b738317e4/recOrder/cli/compute_transfer_function.py#L28
it is actually allowed. What is the indended behavior when no config is given?

recOrder/tests/conftest.py
clinton-huynh: I'm not quite sure what you mean by reorder it to reflect the sequence. Does that mean to make the name something like `settings_compute_transfer_function`?

recOrder/tests/cli_tests/test_cli.py
clinton-huynh: I'll clean up all the unused imports!

recOrder/tests/cli_tests/test_cli.py
```diff
@@ -1,10 +1,154 @@
...
+
+def test_compute_transfer_blank_output():
+    runner = CliRunner()
+    for option in ("-o ", "--output-path "):
+        cmd = "compute-transfer-function " + option
+        result = runner.invoke(cli, cmd)
+        assert result.exit_code == 2
```
ziw-liu, line 47: In the previous test case it is already asserted that no `-c` will cause an error. So this test will pass regardless of the behavior (existence of output arg) it is try to test for.

recOrder/tests/cli_tests/test_cli.py
clinton-huynh: I believe that is for if they add the options "-c" or "--config-path" but do not provide a specific path for the argument. I assumed that the `config_path` being `None` was for having no option and argument, but I didn't realize that it may be for no given config. If we want the command to take in `recorder compute-transfer-function -c ` and still do the default configurations, I can look to make those changes.

recOrder/tests/conftest.py
ziw-liu: Yeah something like `settings_and_transfer_function`.

recOrder/tests/cli_tests/test_cli.py
```diff
@@ -1,10 +1,154 @@
...
+    settings, dataset = setup_default_ctf_settings
+    generate_and_save_birefringence_transfer_function(settings, dataset)
+    assert dataset["intensity_to_stokes_matrix"]
+
+def test_absorption_and_phase_write(setup_default_ctf_settings):
+    settings, dataset = setup_default_ctf_settings
+    generate_and_save_phase_transfer_function(settings, dataset)
+    assert dataset["absorption_transfer_function"]
+    assert dataset["phase_transfer_function"]
+    with pytest.raises(KeyError):
```
ziw-liu, line 80: This is not testing for an error raised from the tested function. To test that the key does not exist in a dict in the local scope simply do:

```py
assert "key1" not in dataset
assert "key2" not in dataset
```

which is much faster.

recOrder/tests/conftest.py
```diff
@@ -0,0 +1,38 @@
+import pytest
+from recOrder.cli.settings import TransferFunctionSettings
+from iohub.ngff import open_ome_zarr
+
+@pytest.fixture(scope="function")
+def setup_default_ctf_settings(tmp_path):
```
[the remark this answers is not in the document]
ziw-liu: Sorry I wasn't clear:

Since each of these fixtures are just used once, the test cases could just all receive the 'default' fixture and modify the settings in themselves.

ziw-liu (approved):
Looks good!

PR #382, opened 2023-06-29
Fix microscope installation guide error

talonchandler:
Fixes a small error @ieivanov and I spotted during a discussion of this work. 

Currently, the next release is planned to include a single `.dll` USB device driver. We might consider distributing this device driver with MM if we get permission from Meadowlark, but this is currently a low priority.

### DIFF
--- a/docs/meadowlark-software-upgrade.md
+++ b/docs/meadowlark-software-upgrade.md
@@ -1,0 +1,8 @@
+# Upgrade Meadownlark PC software version
+
+To upgrade your "PC software version" use these steps:
+
+- From "Add and remove programs", remove CellDrive5000 and "National Instruments Software".
+- From "Device manager", open the "Meadowlark Optics" group, right click `mlousb`, click "Uninstall device", check "Delete the driver software for this device", and click "Uninstall". Uninstall `Meadowlark Optics D5020 LC Driver` following the same steps.
+- Using the USB stick provided by Meadowlark, reinstall the USB drivers and CellDrive5000. 
+- **Upgrading users:** you will need to reinstall the Meadowlark device to your micromanager configuration file, because the device driver's name has changed to `MeadowlarkLC`. 
--- a/docs/microscope-installation-guide.md
+++ b/docs/microscope-installation-guide.md
@@ -47,7 +47,7 @@ Download and install [`Micromanager 2.0` nightly build `20230426` (~150 MB link)
 
 **Note:** We have tested recOrder with `20230426`, but most features will work with newer builds. We recommend testing a minimal installation with `20230426` before testing with a different nightly build or additional device drivers. 
 
-Before launching `Micromanager`, download the USB driver dll from the [release page](https://github.com/mehta-lab/recOrder/releases/) and place this pair of unzipped files into your `Micromanager` folder (likely `C:\Program Files\Micro-Manager` or similar). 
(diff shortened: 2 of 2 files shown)


ieivanov:
From reading https://github.com/micro-manager/mmCoreAndDevices/pull/341 I see that the device adapter which ships with MM is built agains USB driver V108. That USB driver version do we distribute with the GitHub release? The device adapter [instruction page](https://micro-manager.org/MeadowlarkLC) says to use the `.dll` which Meadowlark distributes. Let's point our installation guide to the MM page so we can keep these instruction in one place. Here we could add instructions for legacy installations (if we are supporting those)?

talonchandler:
I'm a bit confused about what we should do with the USB driver. Should we continue to distribute via `recOrder` (the plan that was in my head), or should we stop distributing and tell the user to use the version supplied by Meadowlark (the instructions on the device adapter page)? I'm not sure I know all of the history/permissions with distributing this file. 

> From reading https://github.com/micro-manager/mmCoreAndDevices/pull/341 I see that the device adapter which ships with MM is built against USB driver V108. What USB driver version do we distribute with the GitHub release?

For `recOrder<=0.3.0` we distribute an older USB driver (that works with MM 20210920), and I was planning for the `recOrder==0.4.0` release to be distributed with `usbdrvd.dll` built against V108. Happy to change this plan. 

> The device adapter [instruction page](https://micro-manager.org/MeadowlarkLC) says to use the .dll which Meadowlark distributes. Let's point our installation guide to the MM page so we can keep these instruction in one place.

My (minor?) concern with doing this is that that instruction page isn't pinned to a specific release, so the instructions may change. My intent with these instructions is to keep them specific to the version they're distributed with, so that we have a hope of installing/recreating old versions. 

> Here we could add instructions for legacy installations (if we are supporting those)?

I think we can support legacy version via their legacy documentation. I don't think we should have installation instructions for legacy versions on the main branch.

mattersoflight:
It is great to see the MeadowlarkLC device adapter in the main MM repo! 

My vote is to support only the current version of the device adapter/USB driver (that works with the current version of recOrder) while we are figuring out a software architecture amenable to broader deployment. Do you see a reason to support legacy installations?

ieivanov:
@talonchandler let's chat about this in person

talonchandler:
@ieivanov and I just had a chat about how to handle this:

We decided that when we release recOrder-napari==0.4.0 we will distribute the usbdrvd.dll built against V108 via the recOrder GitHub release page.
We'll link to this driver and instruct the user to do drag-and-drop it into their MM installation. We'll put the instructions in the recOrder docs. It doesn't seem like we'll need to link to the micromanager device driver instruction page.
We decided to keep installation instructions with their versions---no need to maintain legacy documentation. For example, 0.3.0 documentation will live with the 0.3.0 branch/release and 0.4.0 documentation will live with the 0.4.0 branch/release.

ieivanov:
@talonchandler are you still working on this PR or is it ready to merge? Are you planning to add the dependency table here?

talonchandler:
Thanks. I'm finishing up a couple things on this branch right now. I'll request your review on this branch (updated docs in preparation for 0.4.0) and #394 at the same time because I want to keep a couple things synchronized.

talonchandler:
This PR was against `0.4.0dev`, and I couldn't easily wrangle it into a form that would easily merge into `main`. I `cherry-pick`ed the key commits and opened a clear PR for discussion on #395. 

Closing this.

PR #381, opened 2023-06-16
Fluorescence CLI

talonchandler:
@ziw-liu I'm requesting your review of this fairly large PR. I have tested this PR on hummingbird, and I'd like to prioritize getting this merged into `0.4.0dev` then `main` so that we can start integration testing for the `0.4.0` release. 

I'm particularly interested in your suggestions related to (1) the new `/examples/` folder for the new config file formats, (2) the new CLI names and simplified `recOrder reconstruct` CLI, and (3) the new `cli.settings.py` file. 

**Summary of major changes**:
- new fluorescence CLI functionality
- a single configuration file design instead of two configuration files (one for `compute-transfer-function` and one for `apply-inverse-transfer-function`
- reworked pydantic models for configuration files (now supported birefringence, phase, birefringence + phase, and fluorescence)
- reworked `examples/` folder demonstrating new CLI

Minor changes:
- high-level tests of CLI
- new CLI names: `compute-tf` and `apply-inv-tf`. These shortened names are for the CLI only. 
- basic validation of CLI inputs
- automatically generated example configuration files

Earlier list of TODOs:
- [x] finish high-level `recorder reconstruct` CLI and test
- [x] integration testing at the microscope
- [x] removal of example scripts with documentation (now replaced by `waveorder` scripts and CLI + example configurations)
- [x] documentation of configuration file generation

@edyoshikun I'm tagging you as a reviewer, but we'll likely merge before you're back. We'll make sure this is tested against your upcoming use cases before the 0.4.0 release.

PR #383, opened 2023-07-25
Run CI/CD tests on all PRs

talonchandler:
**Before:** Tests run only on PRs to `main`
**After:** Tests run on all PRs 

**Symptom:** In #381 I used some local spot tests, then saw the checkmarks on GH actions and thought everything was working well. On merging #381 into `0.4.0dev`, I found that the tests weren't actually running...the checkmark was only for napari hub checks. Not a big deal; I  will fix the tests in `0.4.0dev`.

### DIFF
--- a/.github/workflows/test_and_deploy.yml
+++ b/.github/workflows/test_and_deploy.yml
@@ -11,7 +11,7 @@ on:
       - "*"
   pull_request:
     branches:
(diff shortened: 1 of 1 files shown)


ziw-liu:
The napari hub preview failed in some TypeScript process...

PR #385, opened 2023-07-25
Remove assertion and unused function

### LINKED ISSUE #213: Stop using `assert` in production code
**Problem**
`recOrder.io.ConfigReader._check_assertions` uses `assert` to validate configs. It is not generally the best practice to let important logics rely on assertions.

**Proposed solution**
We should avoid the usage of assertion when refactoring the config reader for 1.0.0+.

ziw-liu:
Fix #213.

Also remove the orphaned `get_unimodal_threshold` function. This was likely used for segmentation (e.g. finding nuclei mask) and is now part of other analysis pipelines. We can also add a deprecation warning instead of removal for 0.4.0 if anyone is potentially still using it from here.

### DIFF
--- a/recOrder/calib/Calibration.py
+++ b/recOrder/calib/Calibration.py
@@ -106,17 +106,18 @@ class QLIPP_Calibration:
         # Set Mode
         # TODO: make sure LC or TriggerScope are loaded in the respective modes
         allowed_modes = ["MM-Retardance", "MM-Voltage", "DAC"]
-        assert (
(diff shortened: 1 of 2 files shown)


talonchandler:
Thanks!

I looked into the history of `get_unimodal_threshold`, and it seems to be related to some unfinished background subtraction features for fluorescence.

If we want to use this code, I think it's best for it to live outside of `recOrder` so I'm approving.

PR #388, opened 2023-07-26
Live reconstruction compatibility features

talonchandler:
This PR adds two features to make `recOrder`'s CLI reconstruction workflow more compatible with live reconstructions:

1. a new `time_indices` option for reconstructing a subset of all time points. `time_indices` can be an integer, a list of integers, or `all` (default). This allows the user to reconstruct all time points (typical for offline), or individual time points (typical for faster online reconstructions). 

2. support for passing `ReconstructionSettings` objects directly into `apply_inverse_transfer_function_cli`. This feature makes it more convenient for an online reconstruction routine to call the same reconstruction on different time points...instead of modifying and save a configuration yaml file for each time point, the user can modify the `time_indices` attribute of the `ReconstructionSettings` object directly. 

@ziw-liu I'm not sure if I implemented (2) in the way that you had in mind...please suggest improvements. 
@clinton-huynh please feel free to add or modify this branch.

### DIFF
--- a/recOrder/cli/apply_inverse_transfer_function.py
+++ b/recOrder/cli/apply_inverse_transfer_function.py
@@ -19,6 +19,21 @@ from waveorder.models import (
 )
 
 
+def _load_configuration_settings(
+    configuration_settings: Union[str, ReconstructionSettings]
+) -> ReconstructionSettings:
+    if isinstance(configuration_settings, str):
+        return utils.yaml_to_model(
+            configuration_settings, ReconstructionSettings
+        )
+    elif isinstance(configuration_settings, ReconstructionSettings):
+        return configuration_settings
+    else:
+        raise ValueError(
+            f"configuration_settings = {configuration_settings} must be a string or a ReconstructionSettings object."
+        )
+
+
 def _check_background_consistency(background_shape, data_shape):
     data_cyx_shape = (data_shape[1],) + data_shape[3:]
     if background_shape != data_cyx_shape:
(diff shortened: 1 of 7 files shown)


recOrder/cli/apply_inverse_transfer_function.py
```diff
@@ -27,23 +28,55 @@ def _check_background_consistency(background_shape, data_shape):
...
+    """
     echo_headline("Starting reconstruction...")
 
     # Load datasets
     transfer_function_dataset = open_ome_zarr(transfer_function_path)
     input_dataset = open_ome_zarr(input_data_path)
 
-    # Load config file
-    settings = utils.yaml_to_model(config_path, ReconstructionSettings)
+    # Load configuration settings file
```
ziw-liu, line 62: One point we talked about offline is to put this block of code in its own function, since this function is already too long.

recOrder/tests/cli_tests/test_reconstruct.py
```diff
@@ -74,14 +81,25 @@ def test_reconstruct(tmp_path):
...
+            str(tf_path),
+            recon_settings,
+            str(result_path_direct),
+        )
+        assert result_path_direct.exists()
+
         # Check output
         result_dataset = open_ome_zarr(result_path)
-        assert result_dataset["0"].shape[0] == 2
+        assert result_dataset["0"].shape[0] in {1, 2}
```
ziw-liu, line 102:
```suggestion
        t_opt = len(list(time_indices))
        time_dim = t_opt if t_opt >= 2 else 1
        assert result_dataset["0"].shape[0] == time_dim
```

talonchandler:
I'm going to change this to draft because the tests aren't passing and we made some suboptimal choices here to enable a demo. The work in this branch is useful for demonstrating live reconstructions from micromanager, though, so this does not need to be closed until we've found a good solution for that. 

@edyoshikun I am planning to move the `time_indices` option to it's own PR because that feature is useful for multiprocessing.

talonchandler:
Replaced by #398. Closing.

PR #391, opened 2023-08-02
Bump dependencies and enable widget test

ziw-liu:
### Dependencies

- With napari 4.18, Qt6 is now an explicit option, so we can remove the workaround added for #193.
- Bumping iohub to v0.1.0dev4

### Enable GUI test

- With the Qt change, the dock widget test is now passing on Linux CI runners too.


talonchandler (approved):
Thanks @ziw-liu. I tested a fresh install on hummingbird and everything is working well.

Issue #304, opened 2023-01-12
Keep a maintenance branch for each minor version


ziw-liu:
While working on #303 I noticed that there is not a branch to open a PR against, since we always released from `main`. Releasing a stable version without a PR seemed 'dirty', and can be avoided if we kept a maintenance branch for each minor version like [NumPy does](https://github.com/numpy/numpy/branches). I named the #303 branch `patch/0.2.x`, and we can either keep using this format or use another one like `bug-fixes/major.minor.x` or `maintenance/major.minor.x` instead.

talonchandler:
Great idea. I don't have a major preference for the naming---`/patch/0.2.x` makes sense to me.

mattersoflight:
Good idea. I checked if it is possible to branch off a tag and that works as needed. I vote for `maintenance/major.minor.x`. 
`git checkout -b patch/0.2.2 0.2.1`

ziw-liu:
> I checked if it is possible to branch off a tag and that works as needed.

Yes, it should be possible to create a branch off of any commit, including tagged commits, but (I think) only branches can be valid merging targets (because tags are associated with a commit so it cannot be mutated).

talonchandler:
@ziw-liu what needs to happen for this to close? Is this for immediately following the 0.3.0 release?

ziw-liu:
> @ziw-liu what needs to happen for this to close? Is this for immediately following the 0.3.0 release?

I labelled this issue as `meta` since it's about making a decision about repo management. As soon as we all agree that we will (or will not) adopt the strategy, it can be closed.

PR #394, opened 2023-08-03
0.3.1 documentation patch

talonchandler:
This PR implements some of the decisions we made in #382. 
Note that this PR is against `patch/0.3.x`. See #304 for context about keeping maintenance branches. 

After this PR:
- the 0.3.1 documentation will include a compatibility table
- the 0.3.1 documentation will be synchronized with the 0.4.0 documentation in #395 

After this PR merges I will release 0.3.1 and push to PyPI.

### DIFF
--- a/docs/meadowlark-software-upgrade.md
+++ b/docs/meadowlark-software-upgrade.md
@@ -1,8 +1,0 @@
-# Upgrade Meadownlark PC software version
-
-To upgrade your "PC software version" use these steps:
-
-- From "Add and remove programs", remove CellDrive5000 and "National Instruments Software".
-- From "Device manager", open the "Meadowlark Optics" group, right click `mlousb`, click "Uninstall device", check "Delete the driver software for this device", and click "Uninstall". Uninstall `Meadowlark Optics D5020 LC Driver` following the same steps.
-- Using the USB stick provided by Meadowlark, reinstall the USB drivers and CellDrive5000. 
-- **Upgrading users:** you will need to reinstall the Meadowlark device to your micromanager configuration file, because the device driver's name has changed to `MeadowlarkLC`. 
--- a/docs/meadowlark-software-upgrade.md
+++ b/docs/meadowlark-software-upgrade.md
@@ -1,0 +1,9 @@
+# Upgrade Meadownlark PC software version
+
+To upgrade your "PC software version" use these steps:
+
+- From "Add and remove programs", remove CellDrive5000 and "National Instruments Software".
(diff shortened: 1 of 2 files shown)


docs/microscope-installation-guide.md
```diff
@@ -1,11 +1,31 @@
...
-2. Installing a compatible version of `MicroManager` and LC device drivers.
-3. Connecting `recOrder` to `MicroManager` via a `pycromanager` connection.
+1. Checking pre-requisites for compatibility. 
+2. Installing Meadowlark DS5020 and liquid crystals.
+3. Installing and launching the latest stable version of `recOrder` via `pip`. 
+4. Installing a compatible version of `MicroManager` and LC device drivers.
+5. Connecting `recOrder` to `MicroManager` via a `pycromanager` connection.
 
-Before you start you will need a machine with Windows 10, a Meadowlark DS5020 connected to a liquid crystal device, and a microscope system compatible with `Micromanager`. 
+# Compatibility Summary
```
ziw-liu, line 10: Should this be H2 instead of H1?

ziw-liu (approved):

docs/meadowlark-software-upgrade.md
```diff
@@ -0,0 +1,8 @@
+# Upgrade Meadownlark PC software version
+
+To upgrade your "PC software version" use these steps:
+
+- From "Add and remove programs", remove CellDrive5000 and "National Instruments Software".
+- From "Device manager", open the "Meadowlark Optics" group, right click `mlousb`, click "Uninstall device", check "Delete the driver software for this device", and click "Uninstall". Uninstall `Meadowlark Optics D5020 LC Driver` following the same steps.
+- Using the USB stick provided by Meadowlark, reinstall the USB drivers and CellDrive5000. 
+- **Upgrading users:** you will need to reinstall the Meadowlark device to your micromanager configuration file, because the device driver's name has changed to `MeadowlarkLC`.
```
ieivanov, line 8: Does this belong here? In `docs/microscope-installation-guide.md` you say that the device is called `MeadowlarkLcOpenSource`. At what point did we rename it to `MeadowlarkLC`? Also, this is more general MM setup rather than related to upgrading the PC software version, I think

ieivanov: P.S. These are instruction for users upgrading from recOrder 0.3 to recOrder 0.4, I think we can delete the Upgrading users section from here. Or you may wish to leave a comment if anyone wants to downgrade (very unlikely)?

talonchandler: Great calls @ieivanov. I've:
- moved the "PC software version instructions" into the installation guide
- removed the upgrading instructions from here, because they're not needed

ieivanov (approved):

PR #398, opened 2023-08-08
Reconstruct a subset of time indices

talonchandler:
`recorder reconstruct` naively loops through all time points, applying the reconstruction as it goes. This default is limiting for datasets with a large number of time points where it is more convenient to reconstruct a single time point on a single core (via multiprocessing or slurm). 

This PR adds a configuration-file option `time_indices` with signature:
`time_indices: Union[NonNegativeInt, List[NonNegativeInt], Literal["all"]] = "all"`

By default `time_indices: "all"` reconstructs all time point, which is backwards compatible. 
If `time_indices: 5`, then only time-point 5 will be reconstructed.
Finally, if
```
time_indices:
- 0 
- 5
- 10
```
then the 0th, 5th, and 10th indices will be reconstructed. 

@edyoshikun requested this feature for simplifying parallel reconstructions with `recOrder`, and I expect we'll use a similar feature in `mantis`.

### DIFF
--- a/recOrder/cli/apply_inverse_transfer_function.py
+++ b/recOrder/cli/apply_inverse_transfer_function.py
@@ -167,7 +167,7 @@ def apply_inverse_transfer_function_cli(
             transfer_function_dataset["intensity_to_stokes_matrix"][0, 0, 0]
         )
 
-        for output_index, time_index in enumerate(time_indices):
+        for time_index in time_indices:
             # Apply
             reconstructed_parameters = (
                 inplane_oriented_thick_pol3d.apply_inverse_transfer_function(
@@ -180,7 +180,7 @@ def apply_inverse_transfer_function_cli(
             )
             # Save
             for param_index, parameter in enumerate(reconstructed_parameters):
-                output_array[output_index, param_index] = parameter
+                output_array[time_index, param_index] = parameter
 
     # [phase only]
     if recon_phase and (not recon_biref):
@@ -204,7 +204,7 @@ def apply_inverse_transfer_function_cli(
                 transfer_function_dataset["phase_transfer_function"][0, 0]
             )
 
-            for output_index, time_index in enumerate(time_indices):
+            for time_index in time_indices:
                 # Apply
                 (
                     _,
@@ -217,7 +217,7 @@ def apply_inverse_transfer_function_cli(
                 )
 
                 # Save
-                output_array[output_index, -1, 0] = yx_phase
+                output_array[time_index, -1, 0] = yx_phase
 
         # [phase only, 3]
         elif recon_dim == 3:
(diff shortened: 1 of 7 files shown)


ziw-liu:
Should https://github.com/mehta-lab/recOrder/pull/388/commits/d2331d9ad7c63b401956b0da04d5495ace0a62b1 be ported into this PR?

talonchandler:
Great call @ziw-liu. 

I cherry-picked that commit and modified the condition to create a new array (instead of checking if 0 is in time_indices, I'm checking if an array already exists).

recOrder/cli/apply_inverse_transfer_function.py
```diff
@@ -53,8 +54,26 @@ def apply_inverse_transfer_function_cli(
...
+        time_indices = range(input_dataset.data.shape[0])
+        t_shape = input_dataset.data.shape[0]
+    elif isinstance(settings.time_indices, list):
+        time_indices = settings.time_indices
+        t_shape = len(time_indices)
+    elif isinstance(settings.time_indices, int):
+        time_indices = [settings.time_indices]
+        t_shape = 1
+    else:
+        raise ValueError(
```
ziw-liu, line 68: This block will never be reached because pydantic will fail early.

recOrder/cli/apply_inverse_transfer_function.py
```diff
@@ -53,8 +54,26 @@ def apply_inverse_transfer_function_cli(
...
+            f"time_indices = {time_indices} should be `all`, a list of integers, or an integer."
+        )
+
+    # Check for invalid times
+    if np.max(time_indices) > (input_dataset.data.shape[0] - 1):
+        raise ValueError(
+            f"time_indices = {time_indices} includes a time index beyond the maximum index of the dataset = {input_dataset.data.shape[0] - 1}"
+        )
```
ziw-liu, line 76:
```suggestion
    # Check for invalid times
    time_ubound = input_dataset.data.shape[0] - 1
    if np.max(time_indices) > time_ubound:
        raise ValueError(
            f"time_indices = {time_indices} includes a time index beyond the maximum index of the dataset = {time_ubound}"
        )
```

recOrder/cli/apply_inverse_transfer_function.py
```diff
@@ -87,26 +106,31 @@ def apply_inverse_transfer_function_cli(
...
-        name="0",
-        shape=output_shape,
-        dtype=np.float32,
-        chunks=(
-            1,
-            1,
-            1,
+
+    # Create an empty TCZYX array if it doesn't exist
+    if not output_dataset.array_keys():
```
ziw-liu, line 120:
```suggestion
    if "0" not in output_dataset:
```

ziw-liu (approved):

talonchandler:
Thanks for the improvements @ziw-liu!

PR #399, opened 2023-08-09
Pass scale metadata through `recorder reconstruct`

### LINKED ISSUE #397: Passing the voxel (z,y,x) in micrometers in zattrs per position
As I was playing with some datasets for mantis reconstructions using #392, I noticed we don't pass any coordinateTransforms or metadata that can be useful for the rest of the analysis pipelines. It would be useful to pass from the raw zarr the voxel size (z,y,x) dimensions in microns. In fact, these can also be applied upon creation of the zarr stores to load in napari.

talonchandler:
Fixes #397.


ziw-liu (approved):
LGTM!

PR #400, opened 2023-08-09
Remove `recorder view`

talonchandler:
I have replaced my `recorder view` workflows with `iohub`- and `napari-ome-zarr` enabled workflows, so I have not been maintaining `recorder view`.


ziw-liu (approved):
400th 🚀

Issue #396, opened 2023-08-08
`apply_inverse_transfer_function_cli` is not very modular, blocking multiprocessing


talonchandler:
@edyoshikun's attempt to implement multiprocessing in #392 was more challenging than it needed to be because `apply_inverse_transfer_function_cli` is not very modular. 

I would like to take this ~350 line function and break it into smaller functions with:
- one function per reconstruction type: `fluorescence`, `phase`, `birefringence`, `birefringence-and-phase`
- a consistent return shape of `CZYX`
- basic tests for each function
- a `get_reconstruction_output_shape` function (which @edyoshikun has already prototyped in #392)

After a discussion with @edyoshikun, fixing this issue together with a version of #388 will tee up a very simple implementation of multiprocessed reconstructions over timepoints. 

@edyoshikun please comment if you see any holes in this plan.

edyoshikun:
After our discussion last week, I think that this PR is certainly required to have a clean implementation of #392. 

The biggest issue right now is the fact that all reconstructions return different `czyx` shape (i.e birefringence has 3 dim (4,z,y,x), phase and bire 2D return a 2 dim (y,x), and phase and bire 3D return 3 dim (z,y,x)). I think that the waveorder functions are returning the right shapes, but this makes integration with recorder a bit complicated if we want to have a generalized multiprocessing function. Hence, the Frankenstein array indexing and complexities I had to do in #392. 

One solution would be to have individual reconstruction functions that pad the arrays. What I am envisioning is something like the one mentioned above.

```python
def apply_inverse_birefringence(czyx, apply_inverse_bire_args):
      reconstruction_birefringence_zyx = (
              inplane_oriented_thick_pol3d.apply_inverse_transfer_function(
                  czyx, **apply_inverse_bire_args
              )
      ) # this is a tuple of tensors
      #convert tuple of tensor to array
      reconstruction_array = np.array([tensor.numpy() for tensor in reconstruction_birefringence_zyx])
      return reconstruction_array    #(4,z,y,x)
```
In the case of the functions that return a 3D arrays and 2D arrays, I think we can implement a `pad_array_dimensions()` that always returns a 4D array (czyx ). 

To write them into the `output_dataset` we typically slice the `output_dataset[t_idx,c_idx] = reconstruction_array`. Since we are thinking of an approach in #388 where the `t_idx` will be a variable in the configuration, this works well too. 

[...]

PR #408, opened 2023-08-16
CLI creates empty HCS zarr then fills it

### LINKED ISSUE #407: [BUG] `iohub info` does not work with reconstructed hcs plates
After the recent #402 merge, @edyoshikun noticed that `iohub info` does not work at the plate level. 

e.g.
```
iohub info test-output9.zarr/
Reading file:	 /Users/talon.chandler/Desktop/test-data/recOrder/test-output9.zarr
Error: No compatible dataset is found.
```

talonchandler:
Fixes #407, and makes progress towards #396. 

This PR refactors `apply_inverse_transfer_function` so that a complete HCS plate is created in advance then filled. 

@edyoshikun I've changed (and hopefully simplified) some of the logic you were using on the mantis side...see `apply_inverse_transfer_function.py L405-L425`. I'm getting all of the reconstruction-specific metadata, using it to create an empty HCS zarr, then looping through the input positions, inverting and writing as I go. 

I expect/hope that reconstruction-specific metadata can be swapped for deskew-specific or other metadata, and that this pattern is compatible with multiprocessing.

### DIFF
--- a/recOrder/acq/acquisition_workers.py
+++ b/recOrder/acq/acquisition_workers.py
@@ -20,7 +20,7 @@ from recOrder.acq.acq_functions import (
 )
 from recOrder.cli import settings
 from recOrder.cli.apply_inverse_transfer_function import (
-    apply_inverse_transfer_function_cli,
+    apply_inverse_transfer_function_single_position,
 )
 from recOrder.cli.compute_transfer_function import (
     compute_transfer_function_cli,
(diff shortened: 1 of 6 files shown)


edyoshikun (approved):
LGTM! I like the new way to mirror the input to the output. Thanks a lot! @talonchandler

PR #417, opened 2023-08-18
Birefringence overlays on top

### LINKED ISSUE #406: [BUG] Birefringence overlays do not show up on top after repeated acquisitions
After an initial retardance+orientation acquisitions (with an empty layers list) you'll see something like:
image

but subsequent acquisitions result in the birefringence overlay on the bottom:
image

talonchandler:
Fixes #406.

This PR puts a newly generated birefringence overlay as the top layer (for new acquisitions and dragged-in zarr stores). 

The previous logic would only update the birefringence overlay if it did not already exist. This failed to trigger during repeat `recOrder` acquisitions, which write to the birefringence overlay in place so it always exists. 

The new logic only updates the birefringence overlay when the last layer starts with "Orientation", which always follows "Retardance" in the layer order. This logic prevents "mixed updating" bugs where the overlay generation happens with the orientation from one dataset and the retardance from the previous dataset. 

Tested on hummingbird.

### DIFF
--- a/recOrder/plugin/main_widget.py
+++ b/recOrder/plugin/main_widget.py
@@ -1172,30 +1172,29 @@ class MainWidget(QWidget):
 
     def handle_layers_updated(self, event: Event):
         layers: LayerList = event.source
-        latest_layer_name = layers[-1].name
-        channels = ["Retardance", "Orientation"]
-        for ch in channels:
-            if latest_layer_name.startswith(ch):
-                suffix = latest_layer_name.replace(ch, "")
-                other_name = channels[1 - channels.index(ch)] + suffix
+        for layer in layers:
+            if layer.name.startswith("Orientation"):
+                orientation_name = layer.name
+                suffix = orientation_name.replace("Orientation", "")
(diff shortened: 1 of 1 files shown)


ziw-liu:
Does this mean that a dataset with channel order `["Orientation", "Retardance"]` would not trigger the overlay?

talonchandler:
>Does this mean that a dataset with channel order ["Orientation", "Retardance"] would not trigger the overlay?

That's correct, and I don't think we'll encounter the opposite very regularly since we're saving in the same order everywhere. 

If I allowed both orders, then I ran into problems where the overlay would trigger twice.

ziw-liu:
> That's correct, and I don't think we'll encounter the opposite very regularly since we're saving in the same order everywhere.
> 
> If I allowed both orders, then I ran into problems where the overlay would trigger twice.

Then I think this looks good! Just needs documentation somewhere that the channels have to be in this specific order.

talonchandler:
Better docs added! Thanks @ziw-liu.

ziw-liu (approved):

Issue #418, opened 2023-08-18
[BUG] GUI-generated layers are incorrectly scaled


talonchandler:
`iohub convert` and `recorder reconstruct` now pass scale metadata smoothly through their pipelines. 

This introduces a "good" bug: if you drag in a previously acquired zarr store it will be correctly scaled, but the GUI-generated layers are not correctly scaled.

I will add correct scale information for all of the `add_image` functions in `recOrder` to fix this.

edyoshikun:
I saw this bug too, by this you mean you will do something along these lines right?

```python
voxel_size = position.scale
viewer.add_image(image_array, scale=(1,1)+voxel_size)
```

talonchandler:
@edyoshikun see discussion in #419.

The solution will be roughly what you wrote, though there are some subtle edge case when we generate birefringence overlays.

PR #419, opened 2023-08-19
Do not convert scale metadata from micromanager

talonchandler:
Possible fix for #418.

`iohub convert` now passes scale metadata by default, which means that micromanager's pixel sizes get saved into the scale metadata. 

This means that `recOrder` now has two sources of truth for it's scale metadata: the scale metadata of the input zarr and the scale metadata that the user enters in the "pixel size" and "magnification" textboxes. This causes problems because the `ome-zarr` reader treats the zarr's scale metadata as truth, so when you try to compare data from the GUI to data from a file the scale is incorrect: see #418. 

I can see two ways to solve this:
1. **Treat the zarr scale metadata as truth.** This would mean removing the "pixel size" and "magnification" fields from the recOrder GUI, removing the `yx_pixel_size` field from `recOrder` CLI, and forcing all `recOrder==0.4.0` sessions to set the correct pixel size in the micromanager's pixel size calibration tool. This is a reasonably large change. 

2. **Continue to treat the recOrder GUI as truth and do not pass scale metadata through `iohub convert`.** This is the one-line fix implemented here. This means that scales will be [1,1,1,1,1] throughout recOrder sessions. This means that you can easily compare everything, so it "fixes" #418. 

I am leaning towards solution 2 for `recOrder==0.4.0`, and we can use solution 1 for `recOrder==1.0.0`.  

Note: `recorder reconstruct` will continue to pass its input metadata through to the output, so `mantis` can continue to use `recorder reconstruct`.

### DIFF
--- a/recOrder/acq/acquisition_workers.py
+++ b/recOrder/acq/acquisition_workers.py
@@ -354,6 +354,7 @@ class BFAcquisitionWorker(WorkerBase):
                         os.path.join(dir_, prefix),
                         self.latest_out_path,
                         data_type="ometiff",
+                        scale_voxels=False,
                         grid_layout=False,
                     )
                     converter.run()


ziw-liu:
> I will add correct scale information for all of the add_image functions in recOrder to fix this.

This from https://github.com/mehta-lab/recOrder/issues/418#issue-1857408424 seems nicer to me. The old dataset problem can be solved by having a button to set layer scale to identity.
We can still use the GUI scale as the 'source of truth' for `yx_pixel_size`.

talonchandler:
> I will add correct scale information for all of the add_image functions in recOrder to fix this.

Thanks for bumping. I've given this some more thought, and I will add a third option:

3.  **`recOrder` passes scale metadata and uses it for napari visualizations, but `recOrder` does not use scale metadata for reconstructions.** Does you think this is intuitive behavior @ziw-liu? I'm having a tough time deciding between these three options. 

Option 3 will require that all `add_image` calls are scaled using micromanager's metadata that is converted to zarr. This will require emitting the scale from the acquisition threads, but it will work. 

Also, if we choose 1 or 3 I agree that a having a `Scale selected layers to identity` button (or similar) will be easy to implement and useful.

ziw-liu:
> 3\. **`recOrder` passes scale metadata and uses it for napari visualizations, but `recOrder` does not use scale metadata for reconstructions.** Does you think this is intuitive behavior @ziw-liu? I'm having a tough time deciding between these three options.

Is it reasonably easy to compare those 2 scales and raise a warning if they don't match? It might even serve as a reminder that the user forgot to set those GUI values correctly.

talonchandler:
> Is it reasonably easy to compare those 2 scales and raise a warning if they don't match?

I like this idea. I'll try out option 3 with warnings early next week. I'll leave this open for now in case that fails.

talonchandler:
Closing this...#426 merged instead.

PR #420, opened 2023-08-19
Check for PyQt5

### LINKED ISSUE #409: [BUG] If PyQt5 is in your environment, recOrder will use it (causing GUI checkboxes to fail)
@edyoshikun and I tested a `pip install -e .` installation into an existing environment and we ran into Qt bugs. 

We found that the `pip install -e .` did not uninstall PyQt5, so recOrder was using that. 

@ziw-liu is there a way to ensure that a package is not in an environment? If not, we should do fresh conda environments on all of our computers and make sure that PyQt5 is not in our environments.

talonchandler:
Fixes #409.

I tried checking imports like @ziw-liu suggesting, but after `pip uninstall PyQt5` I was still able to `import PyQt5`. The environment variable also failed. 

I tested this solution on hummingbird.

### DIFF
--- a/recOrder/__init__.py
+++ b/recOrder/__init__.py
@@ -1,0 +1,11 @@
+# qtpy defaults to PyQt5/PySide2 which can be present in upgraded environments
+import qtpy
+
+if qtpy.PYQT5:
+    raise RuntimeError(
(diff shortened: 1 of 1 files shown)


ziw-liu (approved):
Good find!

PR #421, opened 2023-08-19
`recOrder` data schema

### LINKED ISSUE #371: [DOC]  Iterate on and document data storage schema
Per our offline discussion, refine and document the data storage scheme for recOrder in YAML format, following this [doc](https://github.com/mehta-lab/microDL/blob/data-storage/micro_dl/data_organization.md).

talonchandler:
Fixes #371 

This PR describes `recOrder`'s current storage schema for discussion. Let's iterate on the schema here, then after this merges I will modify `recOrder` to match the schema. 

Here's a list of the issues I think should be addressed together with my suggestions:
- **consistent numbering:** I'll vote for the first `BG` to be `BG_0` to match the snaps
- **consistent case:** I'll vote for snake_case everywhere, so `BG` -> `bg`, `recOrderPluginSnap` -> `recorder_plugin_snap` and `<acq_name_0>_RawPolDataSnap.zarr` -> `<acq_name_0>_data.zarr`
- **simpler names:** I'll vote for `<acq_name_0>_recOrderPluginSnap` -> `<acq_name_0>_snap`, and `<acq_name_0>_RawPolDataSnap.zarr` -> `<acq_name_0>_raw_data.zarr`
- **consistent path to raw data**: we currently save polarization data to `*_RawPolDataSnap` and brightfield data to `*_RawBFDataSnap`. The channel names within these zarr stores describe the data within (`'BF'`, `'State*'`), so this data is redundant. (I'm flexible on this point, but if we prefer separate names then I think the names should appear one level up...as is they're hidden in the root directory).  
- **consistent duplication of `<acq_name>`**: I will vote for `<acq_name_0>_RawPolDataSnap.zarr` -> `RawPolDataSnap.zarr`, but I'm also fine with the alternative of prepending `<acq_name_0>` to the other files (`reconstruction.zarr`, `transfer_function.zarr`, etc)

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -4,7 +4,7 @@
 
 **I would like to use the `napari plugin`:** start with the [plugin guide](./napari-plugin-guide.md).
 
-**I would like to automate my reconstructions:** start with the [reconstruction guide](./reconstruction-guide.md).
+**I would like to automate my reconstructions:** start with the [reconstruction guide](./reconstruction-guide.md) and use the [data schema](./data-schema.md).
 
 **I would like to watch an experienced user introduce `recOrder` (out of date)**: [watch the tutorial](https://drive.google.com/file/d/1tHLHZ_uUnddb3jzXnQUjUiynSDS8XIy2/view?usp=sharing). 
 
--- a/docs/data-schema.md
+++ b/docs/data-schema.md
@@ -8,15 +8,15 @@ Currently, we structure raw data in the following hierarchy:
 
 ```text
 working_directory/                      # commonly YYYY_MM_DD_exp_name, but not enforced
-├── calibration_metadata_0.txt        
+├── polarization_calibration_0.txt        
 │   ...
-├── calibration_metadata_<i>.txt         # i calibration repeats
+├── polarization_calibration_<i>.txt    # i calibration repeats
 │
 ├── bg_0
 │   ...
 ├── bg_<j>                              # j background repeats
 │   ├── background.zarr             
-│   ├── calibration_metadata.txt        # copied into each bg folder 
+│   ├── polarization_calibration.txt    # copied into each bg folder 
 │   ├── reconstruction.zarr
 │   ├── reconstruction_settings.yml     # for use with `recorder reconstruct`
 │   └── transfer_function.zarr          # for use with `recorder apply-inv-tf`
--- a/docs/data-schema.md
+++ b/docs/data-schema.md
@@ -8,41 +8,37 @@ Currently, we structure raw data in the following hierarchy:
 
 ```text
 working_directory/                      # commonly YYYY_MM_DD_exp_name, but not enforced
-├── calibration_metadata.txt        
+├── calibration_metadata_0.txt        
 │   ...
-├── calibration_metadata<i>.txt         # i calibration repeats
+├── calibration_metadata_<i>.txt         # i calibration repeats
 │
-├── BG
+├── bg_0
 │   ...
-├── BG_<j>                              # j background repeats
+├── bg_<j>                              # j background repeats
 │   ├── background.zarr             
-│   ├── calibration_metadata.txt        # duplicated for each BG
-│   ├── gui_state.yml               
+│   ├── calibration_metadata.txt        # copied into each bg folder 
 │   ├── reconstruction.zarr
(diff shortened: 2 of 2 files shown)


mattersoflight:
Those are good suggestions @talonchandler . We can simplify the schema even further.
* `bg_<n>.zarr` and `<acq_name>_<n>.zarr` can contain raw and reconstructed data as channels. We can even store the YAML configs inside the Zarr store and eliminate one hierarchy level. 
* The transfer functions can be stored as `bg_<n>_tf.zarr` or `<acq_name>_<n>.zarr`.

talonchandler:
Thanks @mattersoflight. I've made edits on this PR for my best-effort attempt to simplify further. 

> bg_<n>.zarr and <acq_name>_<n>.zarr can contain raw and reconstructed data as channels...eliminate one hierarchy level.

This is a challenging change for a couple reasons:
- recOrder's 2D reconstructions have different shapes for raw and reconstructed data, and IMO it would be confusing to mix 2D and 3D data in a single TCZYX array. We could think about adding them as different `Arrays` in the zarr store, but...
- recOrder's current CLI operates on zarrs with regular PTCZYX shapes. Adding array support to recOrder's CLI is possible, but it leads to several non-obvious UI questions: how should I specify which Arrays are input and output (config file or CLI)?, what napari viewer should I use to visualize a multi-channel multi-array zarr store?

At the expense of the extra hierarchy level, I think the current design is working reasonably well so I will vote to stick with separate `raw_data.zarr` and `reconstruction.zarr`. Any thoughts @ziw-liu?

> The calibration metadata, gui_state, and reconstruction_settings can be part of a single YML file, if that does not pose any challenges in workflow.

I think calibration metadata and reconstruction_settings need to stay separate for our workflows (also worthwhile considering an likely upcoming split of `recOrder` and a calibration plugin).

IMO `gui_state` has served its purpose as an interim reproducibility aid since we now have the `reconstruction_settings`. @ziw-liu let me know if you have any objection to me removing it.

ziw-liu:
> * recOrder's 2D reconstructions have different shapes for raw and reconstructed data, and IMO it would be confusing to mix 2D and 3D data in a single TCZYX array. We could think about adding them as different `Arrays` in the zarr store, but...

I agree that the main issue of having 2D reconstructions from 3D raw data in the same store is visualization. The napari-ome-zarr plugin doesn't support extra arrays other than resolution levels. And saving the 2D reconstruction as an extra channel with only Z=0 being filled would render as black for Z!=0, which is not entirely obvious or convenient.

ziw-liu:
> IMO `gui_state` has served its purpose as an interim reproducibility aid since we now have `reconstruction_settings.yml` files. @ziw-liu let me know if you have any objection to me removing it.

It was introduced as a duct-tape fix so no objection in replacing it with a proper implementation!

mattersoflight:
> At the expense of the extra hierarchy level, I think the current design is working reasonably well so I will vote to stick with separate raw_data.zarr and reconstruction.zarr. 

Ah - a good point that reconstruction can be a different shape than raw_data. Makes sense to store `raw_data.zarr` and `reconstruction.zarr`  as separate stores.

> I think calibration metadata and reconstruction_settings need to stay separate for our workflows (also worthwhile considering a likely upcoming split of recOrder and a calibration plugin).
If a CLI needs both config files, the information is best kept in the same files. If a CLI/UI needs just one of the two pieces of information, it can just read that field. The unused field is a 'read-only metadata'. We had too many config files (preprocessing, training, prediction) in microDL that all needed to be carried around. I am trying to prevent a similar outcome.

Is calibration metadata needed only for polarization-resolved acquisitions? If yes, it may make sense to keep the files separate. In that case, I would rename the file to `polarization_calibration.txt`

mattersoflight (approved):

talonchandler:
> Is calibration metadata created only for polarization-resolved acquisitions? If yes, it may make sense to keep the files separate. In that case, I would rename the file to polarization_calibration.txt

That's correct. The calibration is only for polarization-resolved acquisitions, so I like this renaming. Changing. 

The `polarization_calibration.txt` file is not used as an input for the reconstruction pipelines, so I think it's appropriate to keep `polarization_calibration.txt` in a separate file. The only field that is needed by the reconstruction is the swing, so that field appears in both `polarization_calibration.txt` and `reconstruction_settings.yml`. 

Thanks!

PR #423, opened 2023-08-19
Simplify Dask graph before computing overlay

### LINKED ISSUE #416: [BUG] Dragging HCS plates into ome-zarr does not generate overlay
The latest versions of `recOrder` fail to generate birefringence overlays when you drag and drop an HCS plate. 

For now, a workaround is to drag in individual positions. 

Thanks @ziw-liu for the debugging session and for handling the fix.

ziw-liu:
Due to unknown reasons, HCS plates or wells read by ome-zarr-py will be a nested Dask graph. These data objects remain type of `dask.array.core.Array` even after calling `compute()`, which should normally return a NumPy array. This PR calls `compute()` once before computing overlay when it detects such abnormal objects.

Fix #416.

### DIFF
--- a/recOrder/plugin/main_widget.py
+++ b/recOrder/plugin/main_widget.py
@@ -1198,7 +1198,14 @@ class MainWidget(QWidget):
         self, source_names: list[str, str], overlay_name: str
     ):
         def _layer_data(name: str):
-            return self.viewer.layers[name].data
+            data = self.viewer.layers[name].data
+            if isinstance(data, da.Array):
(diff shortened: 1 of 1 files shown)


ziw-liu:
This bug is very nasty... According to Dask documentation the return type of `compute()` should always be the underlying chunk type. Not sure which upstream library is causing it.

talonchandler (approved):
Just tested and it's working great!

Thanks so much for your help here @ziw-liu.

PR #426, opened 2023-08-24
Improved birefringence overlays

### LINKED ISSUE #418: [BUG] GUI-generated layers are incorrectly scaled

talonchandler:
This PR makes the following improvements to the birefringence overlays:
- overlays are correctly scaled, and a warning appears if the input data's scale is different from the reconstruction scale. This fixes #418 and replaces #419.
- overlays are calculated lazily from the most recently added "Retardance*" and "Orientation*" layers via dask arrays. This is enabling for datasets with many timepoints (e.g. recent mantis datasets) where calculating the complete birefringence overlay would crash napari with OOM.
- an "overlay retardance maximum" slider is now on the visualization tab. Lazy computation enables live updating of the overlay. 

The demo gif below demonstrates the main features. Even though I'm only showing a small region, the overlay is being computed for a 2048 x 2048 FOV---the largest we typically use. I estimate the lag time to be 0.1-0.2 seconds, which is noticeable but worth it IMO. You can also "hide" the layer to stop the lazy computation. @ziw-liu and I have discussed plans for 1.0.0 that will allow these computations to happen on a separate thread. 
recOrder-demo-overlays2

Thanks @ziw-liu for your help with this.


ziw-liu (approved):

talonchandler:
Thanks for your help here @ziw-liu!

PR #429, opened 2023-08-25
Background scaling, naming, and squeezing

talonchandler:
This PR makes several QOL improvements to the backgrounds in recOrder:

- correctly passes the scale to the backgrounds (missed in my fixes for #426). Background acquisitions use repeated snaps, so I needed to pass the scale "manually"
- renames the background layers for compatibility with the new overlay generator
- squeezes the arrays before adding them to layers. This fixes a confusing behavior where the background would show raw data with a slider, then subsequent acquisitions would only show a retardance+orientation acquisition when the slider was at position 0.

### DIFF
--- a/recOrder/calib/Calibration.py
+++ b/recOrder/calib/Calibration.py
@@ -9,6 +9,7 @@ import matplotlib.pyplot as plt
 import numpy as np
 from importlib_metadata import version
 from iohub import open_ome_zarr
+from iohub.ngff_meta import TransformationMeta
 from mpl_toolkits.axes_grid1.axes_divider import make_axes_locatable
 from napari.utils.notifications import show_warning
 from scipy.interpolate import interp1d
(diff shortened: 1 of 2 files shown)


recOrder/calib/Calibration.py
```diff
@@ -1090,11 +1091,18 @@ def capture_bg(self, n_avg, directory):
             channel_names=[f"State{i}" for i in range(num_states)],
         ) as dataset:
             position = dataset.create_position("0", "0", "0")
+            from iohub.ngff_meta import TransformationMeta
```
ziw-liu, line 1094: Why is this import delayed?

recOrder/plugin/main_widget.py
```diff
@@ -1154,6 +1154,7 @@ def _add_or_update_image_layer(
         move_to_top : bool, optional
             whether to move the updated layer to the top of layers list, by default True
         """
+        image = np.squeeze(image)
```
ziw-liu, line 1157: Is this supposed to squeeze out the Z dimension?

talonchandler: That's right, and I think it's most intuitive to squeeze everything. Happy to chat.

recOrder/calib/Calibration.py
talonchandler: whoops thanks!

ziw-liu (approved):

PR #438, opened 2023-09-22
slurmkit example for phase3D recon.

edyoshikun:
adding slurmkit examples used for reconstructing the virtual staining pipeline


talonchandler:
Superseded by `submitit`. Thanks @edyoshikun.

PR #441, opened 2023-09-28
Fix bug loading timepoints in `apply_inverse_to_zyx_and_save`

ieivanov:
I think I found a/the bug leading to the large memory footprint of phase reconstructions. `apply_inverse_to_zyx_and_save` would load all timepoints when processing each CZYX volume, which I think is not necessary and requires a lot of time and memory. The code works, but I haven't yet tested if this solves the multiprocessing issue.

### DIFF
--- a/recOrder/cli/utils.py
+++ b/recOrder/cli/utils.py
@@ -51,11 +51,11 @@ def apply_inverse_to_zyx_and_save(
     click.echo(f"Reconstructing t={t_idx}")
 
     # Load data
-    tczyx_uint16_numpy = position.data.oindex[:, channel_indices]
+    czyx_uint16_numpy = position.data.oindex[t_idx, channel_indices]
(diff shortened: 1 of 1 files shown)


talonchandler:
Nice find @ieivanov!

I just benchmarked before and after this PR on a single-time point reconstruction from a 50-time-point dataset:

**Before:**
Job Wall-clock time: 00:00:40
Memory Utilized: 16.12 GB
Memory Efficiency: 1.61% of 1000.00 GB

**After:**
Job Wall-clock time: 00:04:42
Memory Utilized: 94.30 GB
Memory Efficiency: 9.43% of 1000.00 GB

So this was a big source of memory (and time) waste on our most recent runs! Thanks again for digging. 

Naively...I'm surprised that the memory savings was <50x...maybe zarr/numpy plays some tricks so that these types of foot guns aren't maximally bad? I didn't run any benchmarks with multiprocessing.

talonchandler (approved):

PR #442, opened 2023-09-29
POC: GUI layout for recOrder v1.0.0

ziw-liu:
Documenting the v1.0.0 main widget layout in the form of code after discussion with @talonchandler and @mattersoflight. (See changes with `napari -w recOrder-napari` as usual.)

A small complication is that magicgui does not support automatic generation of widgets from pydantic models yet (https://github.com/pyapp-kit/magicgui/issues/474). A workaround would probably look like [this](https://github.com/chrishavlin/miscellaneous_python/blob/master/src/pydantic_magicgui_roundtrip.py).


talonchandler:
Closing this as deprecated. Thanks for the POC @ziw-liu.

PR #446, opened 2023-10-06
Remove the napari hub preview workflow

ziw-liu:
The napari hub preview action is scheduled to be discontinued by upstream, and has been failing in our CI. This PR removes it.


talonchandler (approved):
Thanks!

PR #448, opened 2023-10-10
Dynamic layout of reconstruction wizard

talonchandler:
@naihsuanshao @ziw-liu here's a rough-and-ready version of the dynamic layout of the reconstruction wizard. 

@ziw-liu I still ended up using recursion, though in a modified form. Please suggest/make improvements. 

I used a simple counter to keep track of how many widgets to add/delete on each update. Good idea? Or naive trap?

https://github.com/mehta-lab/recOrder/assets/9554101/cd740db3-cd0a-40c9-83be-1bb48f4f9d44

### DIFF
--- a/recOrder/cli/settings.py
+++ b/recOrder/cli/settings.py
@@ -1,6 +1,5 @@
 import os
 from typing import Literal, Optional, Union
-
 from pydantic import (
     BaseModel,
     Extra,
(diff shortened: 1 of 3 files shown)


ziw-liu:
> I used a simple counter to keep track of how many widgets to add/delete on each update. Good idea? Or naive trap?

If the new widgets are all added to a new container, then we can just hide that one object. (Containers can be nested.)

ziw-liu (approved):

PR #451, opened 2023-10-12
First-pass listen to zarr, reconstruct, and fill napari layers

talonchandler:
This is a rudimentary, single-threaded attempt at our "listen-to-zarr" strategy. 

@ziw-liu I ran into at least three challenges this afternoon/evening:
- when should I create my output zarr, and with what metadata? I ended up using a strategy similar to our earlier `create_empty_zarr`, but now I don't know the number of time points in advance, so I needed to create it for the first time point then append for subsequent time points.  
- how should I use the existing reconstruction API to reconstruct one time point at a time as it arrives? For better or worse, the API mimics the CLI, so the API writes into an existing zarr. Unfortunately, this conflicted with the unknown number of time points, so I have a workaround in this branch where I write the reconstruction to a `tmp.zarr` then copy it into the final .zarr. We may need to rework the API...but I think this suboptimal strategy can work for now. 
- what's the best strategy to create napari layers with a variable/changing size? I couldn't manage to get the napari sliders to update when I appended data to a zarr without removing then replacing the layer. Have you tried this? 

Many of my difficulties stemmed from the unknown number of time points. Clinton assumed that he knew the number of times, right? Would relaxing this constraint make sense for tomorrow?


mattersoflight:
@talonchandler , @ziw-liu I think it is fine to use vanilla zarr or dask if it makes lazy loading of zarr easier than with iohub.

> what's the best strategy to create napari layers with a variable/changing size? I couldn't manage to get the napari sliders to update when I appended data to a zarr without removing then replacing the layer. Have you tried this?
I am not sure if this is the best, but dask seems viable. 
https://napari.org/stable/tutorials/processing/dask.html

ziw-liu (approved):

PR #453, opened 2023-10-12
Pycromanager T and Z sweep acquisition script

talonchandler:
cc @andy-sweet @ziw-liu 

This simple script move the z stage at regular intervals. I've marked the point where `acquire-python` can write a frame to a zarr.


ziw-liu:
@talonchandler can we merge this?

ziw-liu (approved):

PR #454, opened 2023-10-31
Append channels via CLI calls

### LINKED ISSUE #444: Metadata race conditions
Typical `recOrder` reconstructions use globs to select multiple positions like this:
```
recorder apply-inv-tf -i PAL027_DENV_A549_raw.zarr/*/*/* -t phase_tf.zarr -c ./phase.yaml -o test_output.zarr
```
which multiprocesses over positions. This works well, but it doesn't requires a lot of memory on a single node. 

One option is to send indivdual positions to different jobs/processes/nodes with:
```
recorder apply-inv-tf -i PAL027_DENV_A549_raw.zarr/A/1/1/ -t phase_tf.zarr -c ./phase.yaml -o test_output.zarr & 
recorder apply-inv-tf -i PAL027_DENV_A549_raw.zarr/B/2/2/ -t phase_tf.zarr -c ./phase.yaml -o test_output.zarr &
```
but this causes a metadata race condition. Both processes will create a new metadata file because one does not exist, and the second job to complete the metadata creation will "win". For example, if the first process finishes last, `iohub info` will show
```
Row names:		 ['A']
Column names:		 ['1']
```

@ziw-liu and I discussed, and we're strongly considering change `recorder compute-tf` to `recorder prepare`. Instead of just computing the transfer function, `recorder prepare` will take a list of positions and a list of configs (one for each channel) and it will compute the transfer functions and prepare the output zarr store, so that that the parallelizable `recorder apply-inv-tf` can fill. 

This means that `create_empty_zarr` (or a near variant) can be called once in advance, instead of in each process which creates the race conditions.

talonchandler:
**Before this PR:**
This pair of CLI calls
```
recorder reconstruct -i input.zarr/0/0/0 -c ./phase.yaml -o output.zarr
recorder reconstruct -i input.zarr/0/0/0 -c ./DAPI.yaml -o output.zarr
```
would overwrite the first phase reconstruction, resulting in a DAPI-only reconstruction.  

**After this PR:**
The same CLI calls result in two channels in `output.zarr` as expected.


edyoshikun:
This appending channels is also something that I was battling with mantis, especially after the registration as well. I have found that appending is tricky when the process of appending crashes ( for whatever reason, you accidentally close the terminal, end the process,etc) then if you try to restart the appending since the channel name already exists it will fail. I was running both separately to make sure I had successful reconstructions/deskewing and then appending together. Is there a an easier way to cleanup or to re-append?

Soorya19Pradeep:
Thank you @talonchandler for implementing cli call for appending multiple reconstructed channels. I have tested this on the dataset from Hunter in this folder and it is working well. `/hpc/projects/comp.micro/infected_cell_imaging/Image_preprocessing/Exp_2023_09_28_DENV_Fixed/`
The time for processing has reduced to 1/3rd, which is great.
I too have the same question as @edyoshikun. I haven't figured out how to use the same for the dragonfly dataset. I have to reconstruct just the phase channel and append the fluorescence channels. I realize the code already exists in iohub for appending channels. Is there a way to do the same using cli, @ziw-liu ?

edyoshikun:
I also use the iohub code and it's relatively fast.

talonchandler:
Thanks for testing @Soorya19Pradeep!

@Soorya19Pradeep and I just discussed our next steps, and we agreed that:
- there's no need for a "copy-channel" function, since the next downstream step (registration), can operate on two different .zarr stores. 
- this PR is ready for a code review---thanks @ziw-liu 
- when this merges I will release it as `0.4.2rc0`. 
- Phenosight can depend on `0.4.2rc0`, and we're on our way to a one-step installation.

recOrder/cli/apply_inverse_transfer_function.py
```diff
@@ -85,12 +85,14 @@ def apply_inverse_transfer_function_single_position(
...
     output_position_dirpath: Path,
     num_processes,
+    output_channel_names: list[str],
 ) -> None:
     echo_headline("\nStarting reconstruction...")
 
     # Load datasets
     transfer_function_dataset = open_ome_zarr(transfer_function_dirpath)
     input_dataset = open_ome_zarr(input_position_dirpath)
+    output_dataset = open_ome_zarr(output_position_dirpath)
```
ziw-liu, line 95: This variable is redefined at L256.

recOrder/cli/utils.py
```diff
@@ -37,31 +39,44 @@ def create_empty_hcs_zarr(
...
+                position.append_channel(channel_name, resize_arrays=True)
 
 
 def apply_inverse_to_zyx_and_save(
     func,
     position: Position,
     output_path: Path,
-    channel_indices,
+    input_channel_indices,
+    output_channel_indices,
```
ziw-liu, line 60:
```suggestion
    input_channel_indices: list[int],
    output_channel_indices: list[int],
```

recOrder/cli/utils.py
```diff
@@ -37,31 +39,44 @@ def create_empty_hcs_zarr(
                 dtype=dtype,
                 transform=[TransformationMeta(type="scale", scale=scale)],
             )
+        else:
+            position = output_plate[position_key_string]
+
+        # Check if channel_names are already in the store, if not append them
+        for channel_name in channel_names:
```
ziw-liu, line 46: Please document the new channel appending behavior.

recOrder/tests/util_tests/test_create_empty.py
```diff
@@ -31,6 +31,18 @@ def test_create_empty_hcs_zarr():
...
+        chunks,
+        scale,
+        more_channel_names,
+        dtype,
     )
+
+    # Verify existence of appended channel names
+    with open_ome_zarr(store_path, mode="r") as z:
+        for position_key in position_keys:
+            assert "Channel3" in z.channel_names
```
ziw-liu, line 48: The old test (L30-31) was not testing each position, but checking the same plate metadata P times.

ziw-liu: @talonchandler note that I changed C shape in https://github.com/mehta-lab/recOrder/pull/454/commits/ecf1b8a806798ae7fec69ecbbac2e544386c9723.

recOrder/cli/apply_inverse_transfer_function.py
talonchandler: Great catch, fixed.

recOrder/cli/utils.py
talonchandler: Thanks! Documentation improved.

recOrder/tests/util_tests/test_create_empty.py
talonchandler: Great catch!

ziw-liu (approved):

PR #456, opened 2023-11-06
Optional GUI/PM dependencies

ziw-liu:
In PhenoSight and mantis, the pinned Qt and pycro-manager versions makes depending on recOrder difficult.

Here these dependencies are made optional through the `acq` flag.


mattersoflight:
@talonchandler @ieivanov @ziw-liu Since we are now preparing recOrder for use as image analysis library in other projects (mantis, phenosight), it makes sense that napari and pycro-manager are optional dependencies. 

Please review and merge soon. @ziw-liu I suggest that phenosight fetches recOrder from git repo instead of pypi while we fine tune recOrder to work well with it.

cc: @Soorya19Pradeep

talonchandler (approved):
Looks good

mattersoflight (approved):

PR #467, opened 2024-06-04
update iohub dependency

aaronalvarezcz:


ziw-liu:
The CI failure is due to python version mismatch between recOrder (3.9-3.11) and iohub (3.10-3.12).

ziw-liu:
> The CI failure is due to python version mismatch between recOrder (3.9-3.11) and iohub (3.10-3.12).

Potentially depends on https://github.com/mehta-lab/waveorder/pull/168

ziw-liu:
@aaronalvarezcz your commit was merged in #478. Thanks!

PR #473, opened 2024-07-02
Collect and unify color overlays

talonchandler:
Paired with `https://github.com/czbiohub-sf/shrimPy/pull/141`, which will close without merging. 

This PR merges @edyoshikun's color overlays from shrimpy into `recOrder` while making necessary changes to both sets of functions:
- accept `czyx`-shaped arrays as a single input for compatibility with our parallelization strategy. This required a fairly significant change to `recOrder`'s handling of dask arrays. 
- return (3, Z, Y, X)-shaped arrays
- do not have a `channel_order` parameter.

The new function signatures are:
`det ret_ori_overlay(czyx, ret_max: Union[float, Literal["auto"]] = 10, cmap: Literal["JCh", "HSV"] = "JCh"):`
`def ret_ori_phase_overlay(czyx, max_val_V: float = 1.0, max_val_S: float = 1.0):`

I have tested the new `ret_ori_overlay` on the example datasets (to check the lazy-computed dask arrays), and I've tested both new functions in CI. 

**Notes:**
- @edyoshikun, earlier I thought that you'd implemented `ret_ori_phase_overlay` with `JCh`, but I now think that this case is covered by `ret_ori_overlay` with the `JCh` option. 
- I have used this PR to clean up many of the unused pieces of colormapping code across recOrder. All of the visualization now lives in `recOrder.io.visualization`, and I anticipate this code will move to `iphub`.


recOrder/io/visualization.py
```diff
@@ -0,0 +1,164 @@
...
+    if ret_max == "auto":
+        ret_max = np.percentile(np.ravel(retardance), 99.99)
+
+    # Prepare input and output arrays
+    ret_ = np.clip(retardance, 0, ret_max)  # clip and copy
+    # Convert 180 degree range into 360 to match periodicity of hue.
+    ori_ = orientation * 360 / np.pi
+    overlay_final = np.zeros_like(retardance)
+
+    # FIX ME: this binning code leads to artifacts.
```
edyoshikun, line 51: not sure if this is still relevant. otherwise remove these lines

recOrder/io/visualization.py
```diff
@@ -0,0 +1,164 @@
...
+    # levels = 32
+    # ori_binned = (
+    #     np.round(orientation[i] / 180 * levels + 0.5) / levels - 1 / levels
+    # ) # bin orientation into 32 levels.
+    # ori_ = np.interp(ori_binned, (0, 1), (0, 360))
+
+    if cmap == "JCh":
+        noise_level = 1
```
edyoshikun, line 59: should this be exposed as an argument?

recOrder/io/visualization.py
```diff
@@ -0,0 +1,164 @@
...
+    # Normalize the stack
+    ordered_stack = np.stack(
+        (
+            # Normalize the first channel by dividing by pi
+            orientation / np.pi,
+            # Normalize the second channel and rescale intensity
+            rescale_intensity(
+                retardance,
+                in_range=(
+                    np.min(retardance),
```
edyoshikun, line 144: I originally had set it up to do np.min(retardance), but maybe the right thing to have here is to clip with (0,max(retardance)). What do you think?

edyoshikun (commented):
this looks great! This merges the functions as we discusses and tees us up to for the porting to iphub and easier generation of these visualizations.

recOrder/io/visualization.py
talonchandler: Good call. Removed.

recOrder/io/visualization.py
talonchandler: I agree. I moved this to an argument.

recOrder/io/visualization.py
talonchandler: I think the current behavior is an okay start. I think iterating on good default will require testing across a wide range of datasets. I'd like to do this, but I'd vote to keep the scope of this PR tighter.

edyoshikun (approved):
LGTM! Thanks for unifying the implementations @talonchandler

PR #493, opened 2025-01-07
Attempt to fix `pydantic-model-prototype` tests

talonchandler:
This PR is my attempt to fix the tests on the `pydantic-model-protoype` branch. 

@ieivanov demonstrated his local failure to me, and pointed out that napari seemed to hang. Although I'm not able to reproduce @ieivanov's hang locally myself, his demo was a key lead---when I skip `test_dock_widget` (see `d41f4c5`), the tests pass. 

This is a "cheating" fix, though, because this is a useful test. @amitabhverma I now suspect that that your first failing commit `f91b4e9` introduced a regression that is interfering with napari's shutdown/cleanup. I tried a couple easy ideas (see notes below), but I wasn't able to find the core issue. Is it possible that the jobs manager isn't cleaning up correctly? I have not read through every line of `f91b4e9` yet.  

Notes:
`d41f4c5` passes --- this is a "cheating" fix, which skips the napari test. This points me towards hanging processes that are interfering with the napari process.  
`358d514` hangs --- this rules out the cli debugging path as being the issue. 
 `e0e2715` hangs --- this rules out the tab_recon debugging path as being the issue.
 `9b71e19` hangs --- confirming that the HEAD of `pydantic-model-prototype` hangs

### DIFF
--- a/recOrder/cli/main.py
+++ b/recOrder/cli/main.py
@@ -24,6 +24,3 @@ cli.add_command(reconstruct)
 cli.add_command(compute_tf)
 cli.add_command(apply_inv_tf)
 cli.add_command(gui)
-
-if __name__ == '__main__':
-    cli()
--- a/recOrder/cli/main.py
+++ b/recOrder/cli/main.py
@@ -24,3 +24,6 @@ cli.add_command(reconstruct)
 cli.add_command(compute_tf)
 cli.add_command(apply_inv_tf)
 cli.add_command(gui)
+
+if __name__ == "__main__":
+    cli()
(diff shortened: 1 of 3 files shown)


amitabhverma:
Thanks @talonchandler 
This was very useful in finding the issue which seem to arise from the server socket blocking napari closing when running headless for tests. I believe the close event is not triggered in this case which is where I do the cleanup and stopping of threads/socket/processes currently running.

The changes have been incorporated #487

talonchandler:
Great! Closing.

PR #495, opened 2025-01-08
Fix local tests 

### LINKED ISSUE #494: [BUG] Local tests fail when run repeatedly
Local `pytest` once will pass.  
Local `pytest` a second time will fail with:
```
FAILED recOrder/tests/util_tests/test_create_empty.py::test_create_empty_hcs_zarr - assert (1, 3, 1, 1024, 1024) == (1, 2, 1, 1024, 1024)
```

The `test_store.zarr` is not being handled correctly. After `rm -r test_store.zarr`, the tests will pass again.

talonchandler:
Fixes #494. 

Keeps temporary data in `tmp_path`, which gets deleted.


ziw-liu (approved):
Looks good, thanks.

PR #503, opened 2025-02-13
remove pyqt6 dependency and re-worked setup cfgs

amitabhverma:
- removed pyqt6 dependency
- streamline different setup needs to address mehta-lab/waveorder#194
- added position metadata to info label mehta-lab/waveorder#191
- bug report template to include relevant logs which are created next to the output dir, name of the dataset, yaml file(s) if encountering reconstruction errors
- CLI can be used without QtBindings (PyQT or PySide) but will show a warning


recOrder/plugin/__init__.py
```diff
@@ -1,11 +1,16 @@
...
+            "Please remove PyQt5 from your environment with `pip uninstall PyQt5`"
+        )
+    elif qtpy.PYSIDE2:
+        raise RuntimeError(
+            "Please remove PySide2 from your environment with `pip uninstall PySide2`"
+        )
+
+except RuntimeError as error:
+    if type(error).__name__ == "QtBindingsNotFoundError":
+        print("WARNING: QtBindings (PyQT or PySide) was not found for GUI")
```
talonchandler, line 16: Is this still needed?

talonchandler (changes requested):
I tested three installations and all behaved as I expected:
1. `pip install .`. As expected, the CLI worked, the GUI did not (does not appear in CLI list, no napari available). 
2. `pip install .[all]`. As expected, both the CLI and GUI work. 
3. An existing napari installation. `pip install napari[all]` (installs `PyQt5`) then `pip install .[all]`. As expected, both the CLI and GUI work. 

I left a few comments below. Also, please update the README and `/docs` installation instructions to reflect these changes.

recOrder/plugin/__init__.py
talonchandler: Sounds good to me. Making this more general makes a lot of sense.

talonchandler: Thanks for simplifying. Looks good.

talonchandler (approved):
Looks good, thanks @amitabhverma. Just retested the tests I described above, and all's working. Merging.

